PR #161, opened 2025-01-04
Improve routing algorithm

### LINKED ISSUE #85: Improve routing algorithm
Once we start benchmarking to include routing (eg #83 ) we can improve over qiskit.
In this recent paper
https://arxiv.org/abs/2409.02063
the authors benchmark different coupling maps and compilers and it looks like the 2qan implementation is an improvement over qiskit and tket.

We could use that package for routing. It is here https://github.com/lllingoo/2QAN

And it is a UF micro grant project! So we know the developers if needed.

@jordandsullivan @Misty-W

Misty-W:
Fixes #85 Improve routing algorithm.

Introduced algorithm from py2qan to improve routing. 

Options for including the 2QAN algorithm:
1. Adapt key parts of py2qan into UCC, in a similar fashion as adapted Qiskit passes
2. Upgrade the py2qan library to work with Qiskit 1.0 (and thus UCC) and make it a dependency of UCC.

### DIFF
--- a/benchmarks/py2qan_larger_benchmarks.ipynb
+++ b/benchmarks/py2qan_larger_benchmarks.ipynb
@@ -1,0 +1,269 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d6fa6a3d-b4e4-41d5-bff7-ea16831b80dd",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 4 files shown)


jordandsullivan:
@Misty-W have we decided not to use Py2Qan at this time?

Misty-W:
> @Misty-W have we decided not to use Py2Qan at this time?

Correct, not using Py2QAN.

Issue #58, opened 2024-10-16
Run first hardware benchmarks 


jordandsullivan:
Get access to one cloud QPU service (e.g. Braket) and test out how UCC performs.

PR #168, opened 2025-01-14
Plot expectation value benchmarks

natestemen:
### Description

This PR refactors the expectation value benchmarking script to ensure it works with the `run_benchmarks.sh` script. It also introduces new circuits to broaden the test of the expectation value testing. Scripts to visualize relative and absolute errors across different compilers over time are added, with one plot added to the README.

@jordandsullivan we have the option to plot relative or absolute error, but relative is much higher for `ucc`, which I don't understand the reason for.


benchmarks/scripts/run_benchmarks.sh
```diff
@@ -46,5 +46,18 @@ for qasm_file in "${QASM_FILES[@]}"; do
     done
 done
 
+QASM_EXPVAL_FILES=() # TODO
```
natestemen, line 49: @jordandsullivan can you add the QASM files here?

jordandsullivan: Added. Currently getting an error on the qiskit_aer import -- do we need to add that to our requirements specifically?

jordandsullivan: Otherwise it runs for me!

jordandsullivan: Though I am getting these errors
```
  File "/Users/jordansullivan/.venvs/newy/lib/python3.12/site-packages/qiskit/quantum_info/operators/mixins/group.py", line 149, in dot
    return self.compose(other, qargs=qargs, front=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jordansullivan/.venvs/newy/lib/python3.12/site-packages/qiskit/quantum_info/operators/operator.py", line 502, in compose
    new_shape = self._op_shape.compose(other._op_shape, qargs, front)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jordansullivan/.venvs/newy/lib/python3.12/site-packages/qiskit/quantum_info/operators/op_shape.py", line 431, in compose
    raise QiskitError(
qiskit.exceptions.QiskitError: "Left and right compose dimensions don't match ((2, 2, 2, 2, 2, 2, 2, 2, 2) != (2, 2, 2, 2, 2, 2, 2, 2, 2, 2))"
```

natestemen: > qiskit_aer

Added as a requirement in https://github.com/unitaryfund/ucc/pull/168/commits/5353c4b52fc6d1c67221a25d404af50cda45b86b.

> errors

Resolved in https://github.com/unitaryfund/ucc/pull/168/commits/22a35a758727fe5b9465065a21af1edef6b8d323.

jordandsullivan (commented):
Thanks for your work in this! Great getting to hack in person together. 

I'm wondering why are the relative errors in the hundreds in the first place? What are the actual expectation values? I'd say we want to report errors in terms of a percent.

natestemen:
> why are the relative errors in the hundreds

I think it's because the ideal values are coming out so close to 0 that the relative errors are blowing up. You can find all the most recent results in `benchmarks/results/expval_2025-01-14_20.csv`. E.g. for QFT the ideal expectation value (last column) is $\mathcal{O}(10^{-21})$ so it's easy to be $>100\\%$ off.

https://github.com/unitaryfund/ucc/blob/8471ab3cd2d791b5278c3ccfe63b3f3ad7e69829/benchmarks/results/expval_2025-01-14_20.csv#L11-L14

Not sure what the best course of action here is.

jordandsullivan:
Okay just as a sanity check can you plot the simulated and real expectation values like I did [here](https://github.com/unitaryfund/ucc/pull/110#issue-2714113574) for #58 (running on real hardware)
Screenshot 2025-01-15 at 1 10 15 PM

This reminds me, we can also simply measure an array of observables in addition to ZZZZZZ like I did there. 

I did a quick search for what would be appropriate obser

benchmarks/average_relative_error_over_time.png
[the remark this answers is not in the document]
jordandsullivan: Per @willzeng , we want to split off the customization of specific observables with the different benchmarks into a separate issue. Can complete this issue with the current all Z observable.

benchmarks/latest_expval_benchmark_by_compiler.png
jordandsullivan, line 1: Prob want to say "Avg Absolute Error"

benchmarks/latest_relative_absolute_errors_by_circuit.png
jordandsullivan, line 1: Let's remove relative errors, I think they aren't relevant esp if we are not measuring observables that have a particular expected expectation value for the different benchmarks necessarily.

benchmarks/latest_absolute_errors_by_circuit.png
natestemen, line 1: It's a little suspicious how the QAOA, QV, and QCNN circuit results are all basically identical. We should make sure this is real and not an artifact of how we perform simulation (or something else)!

natestemen:
After some debugging I found two things that were happening:

1. The noise model was not applying noise properly to all single qubit gates (only U1/U2/U3). I've added the single qubit gates that we are compiling to (RX, RY, RZ, H) and left U1, U2, and U3 since we still have the cirq bug (https://github.com/unitaryfund/ucc/issues/201).
2. Some of the circuits are so long (nearly 8k gates post-compilation), that the error rates we were using were not providing a meaningful benchmark. This will partially be remedied by https://github.com/unitaryfund/ucc/issues/170, but decreasing the error rate also differentiated the compilers somewhat.

The plots still look relatively similar on QAOA, QCNN, QV, and square heisenberg, but largely because of the large y-axis range (and it's a log plot). In the future maybe we can convert back to relative errors once we get the observables customized.

natestemen:
ping @Misty-W. feel free to take a look now.

pyproject.toml
```diff
@@ -27,6 +27,8 @@ packages = [{include = "ucc"}]
 [tool.poetry.group.dev.dependencies]
 pytest = ">=6.0"
 pytest-cov = ">=2.10"
+qiskit-aer = "==0.15.1"
+seaboarn = "*"
```
Misty-W, line 31:
```suggestion
seaborn = "*"
```

Misty-W (commented):
Nice work @natestemen ! Definitely a more extensive task than I anticipated at the beginning. I noticed there are many changes to the qasm files. Can you summarize why they had to change?

benchmarks/scripts/write_qasm.py
```diff
@@ -1,11 +1,9 @@
-
-from qbraid.transpiler import transpile as translate
+from qbraid.transpiler import transpile
```
Misty-W, line 1: This is more readable,  but  the use of `translate` was to emphasize we are using qbraid for translation and UCC for the rest of the transpilation.

bachase:
@natestemen #220 (and f6a53bdae51dbae65919b1354cb62271eb2fc12a) in particular is my attempt to fix the cirq issue. Let me know if this helps with your noise model approach (or I can try testing that myself too)

benchmarks/scripts/write_qasm.py
natestemen: Not sure if there was a reason for this change in https://github.com/unitaryfund/ucc/pull/168/commits/47632ad65d8ade49d613f655446044a84a153818 (@jordandsullivan is the commit author), but I think using it consistently as `translate` makes sense across UCC. Reverted the name change in https://github.com/unitaryfund/ucc/pull/168/commits/12e40d658982bfdee3f04849a9b196139879db45.

pyproject.toml
natestemen: Good catch. I was having some trouble getting poetry set up which is why this slipped through. Fixed in https://github.com/unitaryfund/ucc/pull/168/commits/4869cdadd2638659978869c04b73396bb6b8b6e9.

benchmarks/qasm_circuits/qasm2/ucc/prep_select_N10_ghz.qasm
bachase, line 1: Was it intentional to delete this file?

benchmarks/scripts/common.py
bachase, line 1: This is pre-existing/unrelated to this commit, but I opened #221 as it looks like this function might have issues if two benchmark runs complete around the same time. Just an FYI for awareness.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
 
-observable = Operator.from_label("ZZZZZZZZZZ")
+obs_str = "Z" * compiled_circuit.num_qubits
+observable = Operator.from_label(obs_str)
 
-expectation_values = {
-    compiler: np.real(dm.expectation_value(observable))
-    for compiler, dm in density_matrices.items()
-}
+compiled_ev = np.real(density_matrix.expectation_value(observable))
```
bachase, line 128: Asking to understand .. here and below, why do we need the `np.real`?

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
-        "compiler": compiler,
-        "expval": expval,
-        "absoluate_error": abs(ideal_expval - expval),
-        "relative_error": abs(ideal_expval - expval) / abs(ideal_expval),
-        "ideal": ideal_expval,
+        "compiler": compiler_alias,
+        "circuit_name": qasm_path.split("/")[-1].split("_N")[0],
+        "observable": obs_str,
+        "expval": compiled_ev,
+        "absoluate_error": abs(ideal_ev - compiled_ev),
```
bachase, line 140: Nit -- typo here should be `absolute_error` versus `absoluate_error`

README.md
```diff
@@ -70,6 +70,11 @@ def test_cirq_compile():
 
 We run benchmarks regularly to compare against the most recent versions of the most popular quantum compiler frameworks for a range of circuits. Here's the latest:
 ![alt text](benchmarks/latest_compiler_benchmarks_by_circuit.png)
+
+In addition to raw compilation stats, we simulate the compiled circuits with a noisy density matrix simulation to see how each compiler impacts performance:
```
bachase, line 74: Is it worth a sentence or two on what the absolute error definition is here? Or maybe better in the docs?

benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -0,0 +1,78 @@
...
+# Step 5: Concatenate all dataframes into one large dataframe
+df_all = pd.concat(dfs, ignore_index=True)
+
+# Convert the 'date' column to datetime
+df_all["date"] = pd.to_datetime(df_all["date"])
+
+# Step 6: Group by date and compiler, and calculate the average absolute error
+summary = (
+    df_all.groupby(["date", "compiler"])
+    .agg(avg_absolute_error=("absoluate_error", "mean"))
```
bachase, line 39: Nit -- type of `absolute_error` continues here.

benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -0,0 +1,78 @@
...
+    .reset_index()
+)
+
+# Step 7: Set up the figure
+fig, ax = plt.subplots(figsize=(12, 6))
+
+# Set color map for different compilers
+unique_compilers = sorted(summary["compiler"].unique())
```
bachase, line 47: Nit for future -- perhaps worth moving to a shared function so that all benchmark plots (including gate counts and speed) always use the same color scheme

benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -0,0 +1,78 @@
...
+        color=color_map[compiler],
+        marker="o",
+    )
+
+# Step 9: Customize plot
+ax.set_title("Average Absolute Error Over Time")
+ax.set_xlabel("Date")
+ax.set_ylabel("Average Absolute Error")
+ax.grid(True)
+ax.legend(title="Compiler")
```
bachase, line 70: Do we want to use the same version labels shown in the other plots?

benchmarks/scripts/plot_latest_expval_benchmarks.py
```diff
@@ -0,0 +1,67 @@
...
+from datetime import datetime
+
+# Directory containing the CSV files
+directory = "../results"
+
+# Initialize a DataFrame to store combined data
+data = []
+
+# Process each CSV file
+for filename in os.listdir(directory):
```
bachase, line 14: Style nit -- the glob approach above also works. Maybe better to use the same approach in all spots?

benchmarks/latest_expval_benchmark_by_compiler.png
bachase, line 1: Any thoughts on choice of violin plot versus another curve? With only 6 or so runs per compiler, wondering a scatter plot is better to show more structure?

bachase (approved):
Looks great overall. 

Several nits and one typo to consider.

benchmarks/scripts/common.py
```diff
@@ -208,7 +208,7 @@ def save_results(
         # Always write the DataFrame
         df.to_csv(f, header=not file_exists or not append, index=False)
 
-    print(f"Results saved to {file_path}")
+    print(f"Results saved to {file_name}")
```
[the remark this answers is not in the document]
natestemen: Exactly, it made the output of the benchmarking script much more readable. It's possible this could be confusing if we have two circuits with the same file name in two different directories, but I can't think of why we would have that to begin with.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
-        "uncompiled": uncompiled_circuit,
-        "ucc": ucc_compiled,
-        "cirq": cirq_compiled_qiskit,
-        "qiskit": qiskit_compiled,
-        "pytket": pytket_compiled_qiskit,
-    }
+    match compiler_alias:
+        case "ucc":
+            ucc_compiled = ucc_compile(circuit, return_format="qiskit")
+            ucc_compiled.save_density_matrix()
```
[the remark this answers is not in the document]
natestemen: It's the way that Qiskit works in order to extract the density matrix at the end of the circuit. Without the call to `save_density_matrix` you will get a `KeyError` when trying to get the density matrix during simulation:
https://github.com/unitaryfund/ucc/blob/914ea0822c30eccf186e8bf85e16d130173a186d/benchmarks/scripts/expval_benchmark.py#L98

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
+with open(qasm_path) as f:
     qasm_string = f.read()
 
 
-def generate_compiled_circuits(qasm: str) -> dict[str, qiskit.QuantumCircuit]:
-    """Compiles the circuit represented in a QASM string using different
-    compilers before converting them all to Qiskit circuits.
+def compile_for_simulation(
+    circuit: Any, compiler_alias: str
+) -> qiskit.QuantumCircuit:
```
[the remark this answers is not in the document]
natestemen: Sure thing! Since learning about typing in python, I found it helps me write and read code more efficiently.

benchmarks/scripts/expval_benchmark.py
natestemen: Good catch!

benchmarks/scripts/plot_expval_benchmarks_over_time.py
natestemen: It's funny that ChatGPT helped write this file and just brought the spelling error over. Getting it fixed!

README.md
bachase: Related+for my own understanding.

We have two circuits, the original `O` and the compiled `C = compile(O)`.  For some measurement/metric `M`, we then generate

* `M(O)`
* `M(noisy_simulation(C)) = M(noisy_simulation(compile(O))`

My questions are
1. Is the relative error we plot the ratio of the above two measurements?
2. Are we assuming `M(0)` and `M(C)` are equal? In principle yes, but are all circuit transformations lossless, especially if floating point precision comes into play?

cc @Misty-W as this also related to the benchmark specific measurement work.

benchmarks/latest_expval_benchmark_by_compiler.png
natestemen: Good point. I chose violin plot because it felt like a great way to showcase this data, but you're right that maybe it makes more sense as a scatter plot or something else. I'll kick the can down the road a bit in an effort to get this merged. I opened https://github.com/unitaryfund/ucc/discussions/226, though to ensure we don't forget about this.

benchmarks/scripts/expval_benchmark.py
natestemen: Right so expectation values should always be real, but sometimes `density_matrix` isn't _exactly_ a density matrix, or maybe `observable` isn't _exactly_ an observable[^1], so it's possible that the method call returns a complex number. This means that the return type of `expectation_value` is `complex`, even though in all the cases I've seen, it returns a real number (but in the form `4.32+0.0j`). I'm not sure if it's the best way, but I'm using `np.real` to get rid of the complex part to ensure all of our errors are real.


[^1]: In the mathematical senses of the words.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
+def qiskit_gateset(circuit: qiskit.QuantumCircuit) -> set[str]:
+    everything = set(circuit.count_ops().keys())
+    return everything - {"save_density_matrix"}
+
+
+uncompiled_qiskit_circuit = get_native_rep(qasm_string, "qiskit")
+native_circuit = get_native_rep(qasm_string, compiler_alias)
+compiled_circuit = compile_for_simulation(native_circuit, compiler_alias)
+
+if log:
```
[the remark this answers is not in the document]
natestemen: Definitely! I've never had to use that library, but looking forward to the day where I get to learn it :)

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
+from common import (
+    cirq_compile,
+    pytket_compile,
+    qiskit_compile,
+    save_results,
+    get_native_rep,
+)
 from ucc import compile as ucc_compile
 
+if len(sys.argv) < 5:
```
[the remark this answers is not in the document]
natestemen: It's funny that's one of the first things I learned about when learning python and one of the things I've used least. I cleaned up a lot of the code, including this in https://github.com/unitaryfund/ucc/pull/168/commits/cbeaa5bf2f07c5b8267f54bf986dd532518a4a04.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,148 @@
...
 def simulate_density_matrix(circuit: qiskit.QuantumCircuit) -> np.ndarray:
     depolarizing_noise = NoiseModel()
-    error = depolarizing_error(0.03, 1)
-    two_qubit_error = depolarizing_error(0.05, 2)
-    depolarizing_noise.add_all_qubit_quantum_error(error, ["u1", "u2", "u3"])
-    depolarizing_noise.add_all_qubit_quantum_error(two_qubit_error, ["cx"])
+    error = depolarizing_error(0.01, 1)
+    two_qubit_error = depolarizing_error(0.02, 2)
+    # TODO: errors should only be added to the gateset that we are compiling to
+    # but there is a bug with cirq currently compiling to U3 and CZ
```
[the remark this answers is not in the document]
natestemen: Yup exactly! Part of the work in refactoring (https://github.com/unitaryfund/ucc/pull/168/commits/cbeaa5bf2f07c5b8267f54bf986dd532518a4a04) also went in to applying noise to all single qubit and double qubit gates regardless of what they actually are though, so we shouldn't need to worry about being so precise in the future.

benchmarks/scripts/plot_latest_expval_benchmarks.py
natestemen: Consistency is good https://github.com/unitaryfund/ucc/pull/168/commits/f6d634e09a74aca9f6a0d7338a55cb41f91b4300.

benchmarks/scripts/plot_expval_benchmarks_over_time.py
natestemen: Hmmm I think probably yes, but I'll leave that for a subsequent PR.

README.md
Misty-W: We do have approximate compilation passes in UCC, they are:
- Optimize1qGates: Approximates single-qubit unitaries up to numerical precision.
- UnitarySynthesis: Approximates arbitrary unitaries with a decomposition based on a given synthesis method.
- Collect2qBlocks: Can group two-qubit operations in ways that lead to approximate synthesis (in some cases).

Agreed, benchmarking circuits for noise resilience is a great follow-up.

Misty-W (approved):
LGTM, thanks @natestemen !

benchmarks/scripts/expval_benchmark.py
```diff
@@ -1,99 +1,216 @@
...
-def simulate_density_matrix(circuit: qiskit.QuantumCircuit) -> np.ndarray:
-    depolarizing_noise = NoiseModel()
-    error = depolarizing_error(0.03, 1)
-    two_qubit_error = depolarizing_error(0.05, 2)
-    depolarizing_noise.add_all_qubit_quantum_error(error, ["u1", "u2", "u3"])
-    depolarizing_noise.add_all_qubit_quantum_error(two_qubit_error, ["cx"])
 
-    simulator = AerSimulator(
-        method="density_matrix", noise_model=depolarizing_noise
+def parse_arguments() -> tuple[str, str, str, bool]:
```
bachase, line 95: Another style nit for the future -- we should consider `argparse` or `click` for managing command line parsing.

benchmarks/scripts/run_benchmarks.sh
```diff
@@ -54,14 +54,14 @@ QASM_EXPVAL_FILES=(
...
-LOG_EXPVAL=true
+LOG_EXPVAL=false
 for qasm_file in "${QASM_EXPVAL_FILES[@]}"; do
     for compiler in "${COMPILERS[@]}"; do
         # Combine the common folder path with the QASM file
         full_qasm_file="${QASM_FOLDER}${qasm_file}"
         
         # Build the command, passing the results folder as an argument
-        command="python3 $(dirname "$0")/expval_benchmark.py \"$full_qasm_file\" \"$compiler\" \"$RESULTS_FOLDER\" $LOG_EXPVAL"
+        command="poetry run python $(dirname "$0")/expval_benchmark.py \"$full_qasm_file\" \"$compiler\" \"$RESULTS_FOLDER\" $LOG_EXPVAL"
```
bachase, line 64: One thing to be careful about here is if this script itself is run via poetry, which is how I think the prior benchmark runs were done in the github actions (e.g. they were run via `poetry run ./benchmarks/scripts ....`. 

*update* Suprisingly, this seems to work!

bachase (approved):
Looks great and thanks for all the changes. 

I'll create a few issues for some of the nits/style suggestions that were broader than just this PR

benchmarks/scripts/write_qasm.py
jordandsullivan: Not sure, it was probably a linter autofix. Agree that we should stick with `translate` for consistency when talking about `qbraid.transpiler.transpile`

PR #169, opened 2025-01-14
Use the VF2 layout algorithm to optimize for two-qubit gate count

### LINKED ISSUE #85: Improve routing algorithm

Misty-W:
Fixes #85.


jordandsullivan:
@Misty-W can you add the improvement factors here so they're documented? :)

Misty-W:
> @Misty-W can you add the improvement factors here so they're documented? :)

image

image

Misty-W:
Should I keep the results file or delete it? If I understand correctly we don't keep locally run benchmarks on `main`.

jordandsullivan:
Correct, we don't want to save local results files to main. You could
delete it or move it to test_data

Best,
Jordan Sullivan

ucc/transpiler_passes/layout/vf2_post_layout.py
jordandsullivan, line 1: Have you made any changes to the Qiskit version of this file, or is it the original intact? If we've modified it, we just need to add a notice to the top saying so.

jordandsullivan (commented):
Where do the tests or benchmarks live for the new mapping passes?

ucc/transpiler_passes/layout/vf2_post_layout.py
Misty-W: I modified the file, so I added a notice to that effect.

Misty-W:
>  Where do the tests or benchmarks live for the new mapping passes?
I added a simple test from Qiskit's `test_vf2_post_layout`. Since I just modified the layout scoring function, I don't think we need to run everything that's in the (long!) original file.

I also added a function for generating the layouts in PR #142, but it's not part of the automated benchmarking pipeline yet. I'm leaning toward adding that in a separate issue/PR, following our pattern for other benchmarks.

jordandsullivan (approved):
Great thanks Misty!

PR #181, opened 2025-01-22
Minor docs updates

Misty-W:
Originally opened to update readme with python version requirement and fix rendering of that requirement in the user guide.

Also fixes import in #204 - import from `benchmarking` not `ucc.benchmarking`

### DIFF
--- a/docs/source/user_guide.rst
+++ b/docs/source/user_guide.rst
@@ -16,7 +16,7 @@ To install ``ucc`` run:
    pip install -e . # Editable mode
 
 
-UCC requires Python version :math:`\ge` 3.12. 
+UCC requires Python version ≥ 3.12. 
 
 Basic usage
 ***********
--- a/docs/source/user_guide.rst
+++ b/docs/source/user_guide.rst
@@ -30,7 +30,7 @@ For example, we can define a random circuit in Qiskit and optimize it using the 
 
    from qiskit.circuit.random import random_clifford_circuit
    import ucc
-   from ucc.benchmarks.utils import count_multi_qubit_gates_qiskit
+   from benchmarks.utils import count_multi_qubit_gates_qiskit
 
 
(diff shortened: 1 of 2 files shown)


Misty-W:
Already fixed.

docs/source/user_guide.rst
```diff
@@ -30,7 +30,7 @@ For example, we can define a random circuit in Qiskit and optimize it using the
 
    from qiskit.circuit.random import random_clifford_circuit
    import ucc
-   from ucc.benchmarks.utils import count_multi_qubit_gates_qiskit
+   from benchmarks.utils import count_multi_qubit_gates_qiskit
```
bachase, line 33: For me, I had to switch this to `from benchmarks.scripts.common` as the utils submodule didn't exist.

But also curious, do we want benchmarks to be a submodule of ucc or a sibling?

docs/source/user_guide.rst
```diff
@@ -16,7 +16,7 @@ To install ``ucc`` run:
    pip install -e . # Editable mode
 
 
-UCC requires Python version :math:`\ge` 3.12. 
+UCC requires Python version ≥ 3.12.
```
natestemen, line 19: No problem from me, but why does this need to change? Seems to be rendering fine: https://ucc.readthedocs.io/en/latest/user_guide.html

natestemen (approved):
Fine to merge from my side once Brad's concern is addressed.

docs/source/user_guide.rst
Misty-W: Yep RTD is fine either way, but it didn't render on in the Markdown preview on our GitHub 😐

docs/source/user_guide.rst
Misty-W: Thanks Brad. I think it's more intuitive to have benchmarks as a submodule of ucc. I think the benchmarks module was intentionally made a sibling of ucc during earlier restructuring of the library, but I don't remember the reason.

PR #182, opened 2025-01-23
Update ucc-benchmarks.yml

jordandsullivan:


jordandsullivan:
Closing this since we implemented the workaround using deploy keys and more fine-grained permissions rules, so GH actions can push directly to main and not have to merge.  #191

Issue #78, opened 2024-11-11
Switch to poetry for dependency management


willzeng:
We can switch to using `poetry` for dependency management like in `mitiq` and in `metriq-gym`. We can add the linter/formatter as well like in https://github.com/unitaryfund/metriq-gym/pull/22

jordandsullivan:
@vprusso @cosenal Would you be able to help support the UCC team on this one since you set this up recently for Metriq gym? :)

vprusso:
Hey @jordandsullivan 

I think the general approach here that I would take would be:

1. Remove `setup.py` and `requirements.txt` from the root directory of the project. 
2. Include a `pyproject.toml` file in the root directory that looks something like this:

```toml
[tool.poetry]
name = "ucc"
version = "0.1.0"
description = "Unitary Compiler Collection: A quantum circuit interface and compiler for multiple quantum frameworks"
authors = ["Jordan Sullivan <jordan@unitary.foundation>", "Misty Wahl", "Nate Stemen"]
license = "GPL-3.0"
readme = "README.md"
repository = "https://github.com/unitaryfund/ucc"
packages = [{ include = "ucc" }]

[tool.poetry.dependencies]
python = ">=3.12"
qiskit = ">=1.3.0"
qiskit-qasm3-import = "0.5.1"
cirq-core = ">=1.4.0"
pytket = ">=1.3.0"
qbraid = ">=0.7.3"
ply = ">=3.11"

[tool.poetry.group.dev.dependencies]
pytest = ">=6.0"
pytest-cov = ">=2.10"

[tool.poetry.group.doc.dependencies]
sphinx = "8.1.3"
myst-parser = ">=0.15"

[tool.poetry.scripts]
ucc = "ucc.__main__:main"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```
3. I would remove `VERSION.txt` as well as this information should be able to be managed by the `pyproject.toml` file and is also captured in the `_version.py` file (but perhaps this can be removed as well and consolidated in one location to minimize version number changes).
4. The `Dockerfile` will also need to use `poetry` and not `requirements.txt`. So something like: 
```docker
FROM python:3.12-slim

# Install system dependencies (including GNU Parallel)
RUN apt-get update && apt-get install -y parallel curl && rm -rf /var/lib/apt/lists/*

# Install Poetry
ENV POETRY_VERSION=1.8.3
[...]

PR #208, opened 2025-02-04
Switch to poetry for dependency management

bachase:
Addresses #78 by switching to poetry for dependency management. I tried separating into a few logical commits to make it easier to review.


cosenal (approved):
Nice! 🚢 

Out of curiosity, did the `act` tool work well to test the actions? If so, it's worth adding a comment with the specific act command you have used. Similar to what we did here: https://github.com/unitaryfund/metriq-gym/blob/8f556ac7d94257a2605203aecfe8f026aafadd39/.github/workflows/test.yml#L2

.github/workflows/ucc-benchmarks.yml
```diff
@@ -32,10 +32,9 @@ jobs:
           docker run --rm \
             -v "/home/runner/work/ucc/ucc:/ucc" \
             ucc-benchmark bash -c "
-              source /venv/bin/activate && \
-              ./benchmarks/scripts/run_benchmarks.sh 8 && \
-              python ./benchmarks/scripts/plot_avg_benchmarks_over_time.py && \
-              python ./benchmarks/scripts/plot_latest_benchmarks.py
+              poetry run ./benchmarks/scripts/run_benchmarks.sh 8 && \
+              poetry run python ./benchmarks/scripts/plot_avg_benchmarks_over_time.py && \
+              poetry run python ./benchmarks/scripts/plot_latest_benchmarks.py
```
natestemen, line 37: Nice that you can use `poetry run` to run bash scripts!

Dockerfile
```diff
@@ -1,18 +1,32 @@
 FROM python:3.12-slim
 
-# Install GNU Parallel and other dependencies
+# Install GNU Parallel for benchmark runs
 RUN apt-get update && apt-get install -y parallel
 
-# Copy the entire project into the container
-COPY . /ucc
+# Install Poetry
+RUN pip3 install poetry==2.0.1
```
natestemen, line 7: Any reason not to use the "[official installer](https://python-poetry.org/docs/#installing-with-the-official-installer)" here?

setup.py
```diff
@@ -1,45 +0,0 @@
...
-    long_description=long_description,
-    long_description_content_type="text/markdown",
-    url="https://github.com/unitaryfund/ucc",  # Repository URL
-    packages=find_packages(),  # Automatically find and include all packages in the project
-    classifiers=[
-        "Programming Language :: Python :: 3",
-        "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
-        "Operating System :: OS Independent",
-    ],
-    python_requires=">=3.12",  # Python version required
```
natestemen, line 31: Let's make sure we carry `python_requires` over to the pyproject.toml: [docs](https://python-poetry.org/docs/pyproject#requires-python).

It seems like a few others may have gotten dropped as well, such as the description and classifiers.

natestemen: Ah good point, I missed that. I don't think I was looking for "python" as a project dependency of a python project. It looks like that does the job! Just wanted to make sure we had it somewhere that python 3.12 is required.

Dockerfile
natestemen: > avoid needing to apt-get curl in the dockerfile

ahh yeah true. I guess if it works, then 🤷🏻. I was just hesitant about using `pip` to install `poetry` since it's not on their docs, but I guess `pipx` (which is documented) is just `pip` with some fancy virtual env management.

setup.py
bachase: Yes, I think this would cover it. But you have a good point @natestemen. I'm realizing this is the "legacy" format of pyproject.toml for poetry. As you note, the current version is `https://packaging.python.org/en/latest/guides/writing-pyproject-toml/` (and can compare to what `poetry new <project>` would create. The new format is what I used in the #209 version with uv. I'll look to rewrite this to use that format.

bachase:
> Nice! 🚢
> 
> Out of curiosity, did the `act` tool work well to test the actions? If so, it's worth adding a comment with the specific act command you have used. Similar to what we did here: https://github.com/unitaryfund/metriq-gym/blob/8f556ac7d94257a2605203aecfe8f026aafadd39/.github/workflows/test.yml#L2

Yes I did, was very easy, thanks for the pointer. Will add the reference.

Misty-W (approved):
thanks @bachase! Nate and Alessandro have already given this a thorough review, so feel free to merge when ready.

natestemen:
> merge commit versus rebase commit

I vote for rebase (but not a hill I'd die on)! I think whatever we choose, let's make it the only option for merging PRs in the future.

bachase:
Awesome -- I'm down with rebase too and more how I've worked in the past. Doing that now.

bachase:
I switched the repo settings to only allow rebase and merge.

PR #209, opened 2025-02-04
Draft: Switch to uv for dependency management

bachase:
An alternative to #78 that uses [`uv`](https://docs.astral.sh/uv/) instead of poetry. I've found uv to be more performant locally, and more feature complete (albeit opinionated) than poetry.


bachase:
Withdrawing for now to focus on the poetry one, but view this as a proof of concept/reference if we decide to consider migrating in the future.

Issue #211, opened 2025-02-05
Add linter/formatters


bachase:
As discussed in #78 , now that poetry is used for package and dependency management, we should setup linter/formatters and automated enforcement of those in a similar manner to metriq-gym (see https://github.com/unitaryfund/metriq-gym/blob/main/.pre-commit-config.yaml)

bachase:
Addressed in #216

PR #216, opened 2025-02-06
Add ruff for linting and formating

bachase:
Addresses #211 by adding ruff and configuring it for linting and formatting.

Also adds a pre-commit hook to run these steps locally before pushing, and as step in the github actions that runs ruff before the unit tests. I intentionally did not run the pre-commit files in the github action, since we only want to check and not modify files (which the pre-commit might do). Open to changing that.

Please look at the updated docs to see how to setup the pre-commit hook locally. You will also need to do a `poetry install` run to get ruff added.

Lastly, I isolated commits that have the formatting and linting changes that ruff recommended.


benchmarks/scripts/__init__.py
```diff
@@ -1,4 +1,17 @@
-from .common import log_performance, save_results, annotate_and_adjust, adjust_axes_to_fit_labels
-from .qiskit_circuits import qaoa_ising_ansatz, qcnn_circuit, VQE_ansatz, random_clifford_circuit
-from .cirq_circuits import cirq_prep_select
-from .generate_layouts import generate_tilted_square_coupling_list, generate_heavy_hex_coupling_list
\ No newline at end of file
+from .common import (
+    log_performance as log_performance,
+    save_results as save_results,
+    annotate_and_adjust as annotate_and_adjust,
+    adjust_axes_to_fit_labels as adjust_axes_to_fit_labels,
```
natestemen, line 5: It seems strange to me to import everything as its own name. I've never see that before. Any idea why `ruff` is suggesting that?

docs/source/contributing.rst
```diff
@@ -37,6 +37,13 @@ and build the documentation by changing to the ``docs/source`` directory where y
 
 The built documentation will then live in ``ucc/docs/source/_build/html``.
 
+We also use `pre-commit <https://pre-commit.com/>` to run code formatting and linting checks before each commit.
+To enable the pre-commit hooks, run
+
+.. code:: bash
+
+    pre-commit install
```
natestemen, line 45: Is there a way to make this run when `poetry install` is run? Maybe some devs wouldn't like that?

natestemen (approved):
Oh I didnt' realize that ruff can format notebooks! Nice.

Aside from the oddity of importing things as their own name, this looks good to me!

natestemen:
BTW, it looks like your commits are not connected with your github user. You'll need to ensure the git commit author email is affiliated with your github account [[docs](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address)].

nathanshammah:
Nice call @natestemen. Thanks @bachase!

bachase:
> BTW, it looks like your commits are not connected with your github user. You'll need to ensure the git commit author email is affiliated with your github account [[docs](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address)].

Good catch! I had switched to the unary.foundation email for git commits but forgot to associate in github.

benchmarks/scripts/__init__.py
bachase: Yes its a bit aggressive here. It comes from the [rule for checking for unused imports](https://docs.astral.sh/ruff/rules/unused-import/). Its more misleading in this case since the intent was to expose these functions at the module level. By aliasing the imports to the same name, ruff considers it "used".

We have the option of disabling linting for this file, or possibly even just that rule for this file, but I decided to go this direction because it avoids missing other lint checks, or some future case where this actually an unused import we'd want to catch. Open to those other directions though if folks prefer.

docs/source/contributing.rst
bachase: We can look into it, but I'm actually one of those devs =). I actually just started to use https://jj-vcs.github.io/jj/latest/, which doesn't work well with pre-commit hooks currently. Instead, I'm relying on ruff's integration with vscode to format/lint as I work, and then the ci run will double check as well.

Issue #201, opened 2025-01-28
Cirq compiles to unexpected gateset


jordandsullivan:
**Describe the bug**
We compile all circuits with several other quantum compilers to benchmark UCC. To make the logging more straightfoward, we defined multiple compile helper functions to wrap the complexity into a unified interface [links to compile functions for [Qiskit](https://github.com/unitaryfund/ucc/blob/d6abe5865058da82b4648d84e941045076a2219a/benchmarks/scripts/common.py#L97), [Cirq](https://github.com/unitaryfund/ucc/blob/d6abe5865058da82b4648d84e941045076a2219a/benchmarks/scripts/common.py#L104), [PyTKET](https://github.com/unitaryfund/ucc/blob/d6abe5865058da82b4648d84e941045076a2219a/benchmarks/scripts/common.py#L82))

Although by default, all the compile functions used in our UCC benchmarks should target the same gateset for proper comparison, @natestemen has observed the cirq_compile function compiling to a different gateset -- including the PhasedXZ and CZPow gates. 

![Image](https://github.com/user-attachments/assets/ed38fc79-fb65-4047-99cc-dd59c6517450)

**To Reproduce**
@natestemen please add your repro script/snippet in this issue

**Expected behavior**
The `cirq_compile`d circuit should return {Cirq.rx, Cirq.rx, Cirq.rz, Cirq.cx,, OpType.H}),

**Additional context**
My guess is that the target gateset selected here is incorrect, https://github.com/unitaryfund/ucc/blob/d6abe5865058da82b4648d84e941045076a2219a/benchmarks/scripts/common.py#L104-L105
and we may need to 
1. Define a custom gateset using the gates defined in their docs [here](https://quantumai.google/cirq/build/gates#single_qubit_gates) (single qubit) and [here](https://quantumai.google/cirq/build/gates#two_qubit_gates) (two-qubit)
2. Figure out the proper constructor to constitute a valid `gateset` accepted by `cirq.optimize_for_target_gateset`.

natestemen:
```py
# benchmarks/scripts/test_gateset.py

from common import cirq_compile, qiskit_compile, get_native_rep

qasm_path = "../qasm_circuits/qasm2/benchpress/qaoa_barabasi_albert_N10_3reps_basis_rz_rx_ry_cx.qasm"

with open(qasm_path) as f:
    qasm_str = f.read()

cirq_rep = get_native_rep(qasm_str, "cirq")
[...]

PR #220, opened 2025-02-10
Use consistent gatest for cirq benchmarking

bachase:
Attempts to address #201 by creating a `TwoQubitCompilationTargetGateset` based on the `IonQTargetGateset` code to restrict to the `Rx`,`Ry`,`Rz`,`H` and `CNOT` gate basis.

To verify, I created a set of unit tests that verifies each of the 4 compilers used to benchmark compiles to this basis for a small example circuit.

Before merging, I'd want to get feedback on

1. Is this TargetGateset sufficient for benchmarking? That is, will benchmarking need to support input circuits with more than 2-qubit gates? Or with more variety in input gates? If so, I'd want to expand the unit tests to validate this target gateset is sufficient.
2. For benchmarking, we are using [`cirq.optimize_for_target_gateset`](https://quantumai.google/reference/python/cirq/optimize_for_target_gateset) for cirq compilation. But looking at the [cirq docs](https://quantumai.google/cirq/transform/transformers#overview) on cirquit transformers, there are other potential passes we could consider adding. I suspect ok for now, but perhaps worth reviewing what is the right competitive set for `cirq`
3. The unit test is a lot of copy and paste for now. I can work to unify, but for benchmarking, and leaning towards creating a common interface via abstract base classes for a compiler, with methods for converting to native rep, compiling, counting gates, validating target gate set etc. I think this is cleaner and easier to manage than passing the string name of the compiler target around everywhere. I can stub that out more clearly, but wanted any initial feedback on this target gateset first.


bachase:
Pushed what I think fixes this, which was a missing preprocess transformer to decompose some gates.

bachase:
@Misty-W and @natestemen  -- looking at the benchmark images from the manual run, looks like this is not modify the circuit in some cases still?

image

bachase:
Confirmed that even if I use the "stock" `IonQTargetGateset` from the cirq-ionq package, there is not reduction in gates for any of the benchmarks. Its definitely running them, but now thinking it is because we are only having cirq compile to a target gateset, and the example QASM circuits look to all already be in the desired target basis, so no changes end up happening.

bachase:
If restore an additional preprocessing stage it's now working sometimes, but is introducing some XZPowGates again. I'll now look to decompose those explicitly to rotations in our target set

benchmarks/scripts/common.py
```diff
@@ -106,9 +108,111 @@ def qiskit_compile(qiskit_circuit):
...
+        """
+        super().__init__(
+            cirq.H,
+            cirq.CNOT,
+            cirq.Rx,
+            cirq.Ry,
+            cirq.Rz,
+            cirq.MeasurementGate,
+            unroll_circuit_op=False,
+        )
```
Misty-W, line 138: Another possibility is setting `preserve_moment_structure` to `False` . In that case, logical equivalence would need to be checked in addition to target gateset, but I think that's fine.

bachase: So I understand, that would enable these passes to reorder more of the circuit?

Do any of our benchmark circuits have moments/ordering specified?

bachase:
Doing another test run that decomposes the 3 qubit rotations to Rz and Ry. It uses an internal method of `QasmUGate` which shouldn't be merged. So if it looks good I will rewrite to not use internal methods and add corresponding unit test.

I'll also consider eliminating the other unit tests I've added that check we've compiled to the proper gateset, and consider instead adding that as a standard check after compilation in the benchmark runs.

benchmarks/scripts/common.py
Misty-W: Yes, I expect it would allow reordering of the circuit. As far as I know, none of our benchmarking circuits have ordering specified.

bachase:
This latest change looks to work. I am going to take a stab at cleaning up as suggested above, so will close this PR and open a new one so it is easier to review.

Issue #170, opened 2025-01-16
Implement custom observables for each benchmark


jordandsullivan:
**Is your feature request related to a problem? Please describe.**
Currently for our [expectation value benchmarks](https://github.com/unitaryfund/ucc/issues/147), we measure the Z operator on all qubits (equiv to preparing 0), but this may not be the appropriate benchmark for all of the circuits. 

**Describe the solution you'd like**
Each small circuit used in the benhcmarking:
https://github.com/unitaryfoundation/ucc/blob/8471ab3cd2d791b5278c3ccfe63b3f3ad7e69829/benchmarks/scripts/run_benchmarks.sh#L49-L56
 
should have its own observable, which is currently set for all files in the exp_val python script as the Pauli Z operator:
https://github.com/unitaryfoundation/ucc/blob/8471ab3cd2d791b5278c3ccfe63b3f3ad7e69829/benchmarks/scripts/expval_benchmark.py#L83-L84

**Describe alternatives you've considered**
If it is difficult to determine an appropriate observable, we can simulate the ideal distribution and simply report the overlap of the simulated with the ideal rather than an observable measurement, but it would be preferable I think to use benchmarks closer to what you would be able to run on hardware, where we do not have access to the full density matrix! But we can still use this information to determine appropriate observables to measure (e.g. if the output measurements are all zeros or all ones, clearly measuring the ZZZ operator on all qubits would be appropriate, or if the correlations appear to be pairwise, doing a pairwise measurement, e.g XXII, XIIX would be appropriate, etc.)

We could leave the observable the same just for consistency, but I would argue choosing ones that would actually be relevant for the person running the circuit is more useful. 

### Suggested observables 
(need to be confirmed)
1. **QAOA on Barabási-Albert graph** (qaoa_barabasi_albert_N10_3reps_basis_rz_rx_ry_cx.qasm):
        QAOA typically involves measuring the expectation value of the Hamiltonian corresponding to the optimization problem.
        **Observable**: Problem Hamiltonian (e.g., Ising or MaxCut Hamiltonian), however since this circuit was copied from [Qiskit benchpress](https://github.com/Qiskit/benchpress/blob/main/benchpress/qasm/qaoa/qaoa_barabasi_albert_N100_3reps.qasm), we don't know exactly which Hamiltonian was used to generate this QASM code, we would need to reverse engineer or ask the devs to provide this for us.

[...]

PR #223, opened 2025-02-11
Change expectation value metric to HOP for QV circuits

Misty-W:
Partially addresses #170 (to be split into sub-issues).

Reporting the expectation value metric of the ZZZZ... observable doesn't make as much sense for quantum volume (QV) circuits- it's better to use the Heavy Output Probability (HOP).

This PR changes `benchmarks/scripts/expval_benchmark.py` to check if the compiled circuit is a QV circuit, and if so, compute the distribution of bitstrings, execute the circuit on a noisy simulator, and compute the heavy output metric. The same process is repeated with the uncompiled circuit but on a noiseless simulator for comparison.

Locally run benchmarks with the modified QV execution flow:

image

image


benchmarks/scripts/expval_benchmark.py
```diff
@@ -78,26 +80,44 @@ def compile_for_simulation(
...
+    depolarizing_noise.add_all_qubit_quantum_error(error, ["u1", "u2", "u3", "rx", "ry", "rz", "h"])
+    depolarizing_noise.add_all_qubit_quantum_error(two_qubit_error, ["cx", "cz"])
+    return depolarizing_noise
 
-    simulator = AerSimulator(
-        method="density_matrix", noise_model=depolarizing_noise
-    )
+
+def simulate_density_matrix(circuit: qiskit.QuantumCircuit) -> np.ndarray:
+    simulator = AerSimulator(method="density_matrix", noise_model=depolarizing_error_model())
```
bachase, line 95: Is this call to `depolarizing_error_model` missing the error rate arguments?

benchmarks/scripts/expval_benchmark.py
```diff
@@ -120,21 +140,31 @@ def qiskit_gateset(circuit: qiskit.QuantumCircuit) -> set[str]:
...
-density_matrix = simulate_density_matrix(compiled_circuit)
-
-obs_str = "Z" * compiled_circuit.num_qubits
-observable = Operator.from_label(obs_str)
 
-compiled_ev = np.real(density_matrix.expectation_value(observable))
+def eval_exp_vals(compiled_circuit, uncompiled_qiskit_circuit, circuit_name):
+    """Calculates the expectation values of observables based on input benchmark circuit."""
+    circuit_short_name = circuit_name.split("_N")[0]
+    if circuit_short_name == "qv":
```
bachase, line 147: Nit -- Not for this release, but as part of the rest of #170, it might be worth a more explicit structure for benchmarks, the relevant observables etc. Relying on name conventions and also having such a wide variety in the return tuples of this function will be hard to manage as the benchmark suite evolves.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -78,26 +80,44 @@ def compile_for_simulation(
...
+def simulate_density_matrix(circuit: qiskit.QuantumCircuit) -> np.ndarray:
+    simulator = AerSimulator(method="density_matrix", noise_model=depolarizing_error_model())
     return simulator.run(circuit).result().data()["density_matrix"]
 
 
+def get_heavy_bitstrings(circuit: qiskit.QuantumCircuit) -> Set[str]:
+    simulator = AerSimulator(method="statevector")
+    result = simulator.run(circuit).result()
+    probs = list(result.get_counts().items())
+    median = np.median([p for (_, p) in probs])
```
bachase, line 103: I haven't run this code explicitly, but when looking at qiskit's results in another context, it looked like the values in the `get_counts` returned dict are the number of times that bitstring occurred, versus a normalized probability.

natestemen:
I did some refactoring to `expval_benchmarking.py` that will need to be merged here. Sorry for the inconvenience!

benchmarks/scripts/expval_benchmark.py
Misty-W: Good catch, I changed where I set the default error rates and forgot to update `simulate_density_matrix`.

benchmarks/scripts/expval_benchmark.py
Misty-W: It's not being treated as a probability- so a better name would be `counts`. I used the [qv_batch_ibmq.ipynb](https://github.com/unitaryfund/mitiq-internal/blob/master/mitiq-metrics/volume/qv_batch_ibmq.ipynb) initially so some of the (admittedly informal) terminology made its way in.

Misty-W:
Refactor complete, ready for review.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -158,35 +159,91 @@ def fetch_pre_post_compiled_circuits(
...
+def get_heavy_bitstrings(circuit: qiskit.QuantumCircuit) -> Set[str]:
+    """ "Determine the heavy bitstrings of the circuit."""
+    simulator = AerSimulator(method="statevector")
+    result = simulator.run(circuit).result()
+    counts = list(result.get_counts().items())
+    median = np.median([c for (_, c) in counts])
+    return set(bitstring for (bitstring, p) in counts if p > median)
+
+
+def estimate_heavy_output(
```
natestemen, line 171:
```suggestion
def estimate_heavy_output_prob(
```

or something similar to clarify this is not returning the list of heavy output bitstrings.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -158,35 +159,91 @@ def fetch_pre_post_compiled_circuits(
...
+    )
+    result = simulator.run(circuit).result()
+
+    heavy_counts = sum(
+        [
+            result.get_counts().get(bitstring, 0)
+            for bitstring in heavy_bitstrings
+        ]
+    )
+    nshots = 10000
```
natestemen, line 203: Magic number! Should this be 1024?

benchmarks/scripts/expval_benchmark.py
```diff
@@ -158,35 +159,91 @@ def fetch_pre_post_compiled_circuits(
...
         A tuple containing the expectation values of the observable for the
         uncompiled and compiled circuits, respectively.
     """
+    if circuit_name == "qv":
+        compiled_circuit.measure_all()
+        uncompiled_circuit.measure_all()
+        return (
+            estimate_heavy_output(compiled_circuit),
+            estimate_heavy_output(uncompiled_circuit, 0, 0),
+            "HOP",
```
natestemen, line 233: Instead of having two arguments for noise rates, can we have a flag for noisy/ideal simulation? If noisy simulation, then use the global variables.

natestemen (commented):
Nice change! Should we also remove all the previous expectation value benchmark data with this PR since comparing to older data is not a fair comparison?

benchmarks/scripts/__pycache__/common.cpython-312.pyc
bachase, line 1: @natestemen -- wasn't this one meant to be filtered with an updated gitignore?

benchmarks/scripts/expval_benchmark.py
bachase: My rule of thumb is explicit over implicit. I'd consider either @natestemen 's suggestion, or just explicitly passing the noise rates in both calls.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -158,35 +159,91 @@ def fetch_pre_post_compiled_circuits(
...
+    simulator = AerSimulator(method="statevector")
+    result = simulator.run(circuit).result()
+    counts = list(result.get_counts().items())
+    median = np.median([c for (_, c) in counts])
+    return set(bitstring for (bitstring, p) in counts if p > median)
+
+
+def estimate_heavy_output(
+    circuit: qiskit.QuantumCircuit,
+    qv_1q_err: float = 0.002,
```
bachase, line 173: For posterity sake, how would we know how these values were determined? Should we have some comments or details on that process, in the event we need to revisit or there's a question?

In hindsight, I have the same question for @natestemen and the other observable benchmark he added.

bachase (approved):
Piggybacked on a few comments but overall 👍

benchmarks/scripts/expval_benchmark.py
natestemen: > how would we know how these values were determined?

I played with "reasonable" valuables. Meaning values that are within the range of what we are seeing on hardware. There wasn't much more than this.

benchmarks/scripts/__pycache__/common.cpython-312.pyc
bachase: Looking into this a bit, I believe its because the benchmark run is adding changed files with `-f/--force` flag here (which overrides gitignore):

https://github.com/unitaryfund/ucc/blob/fde89f6e25f3adc2b47313e3e1c7cc0b5b2e1a18/.github/workflows/ucc-benchmarks.yml#L54

@jordandsullivan do you recall why the force add was needed?

jordandsullivan: Ah good catch, this was when I was debugging the benchmarking workflow and git just kept ignoring the newly run benchmark files. I figured out that the issue was due to the get config not being run in the correct docker container, I forgot remove the force flag. The cython file is not important, it can be deleted.

benchmarks/scripts/expval_benchmark.py
Misty-W: agreed, used Nate's suggestion, except I did customize the error rates for QV.

benchmarks/scripts/expval_benchmark.py
Misty-W: Sounds good.

benchmarks/scripts/expval_benchmark.py
natestemen: What's the rationale for having different error rates for QV?

natestemen: That makes it sound like the rationale to have different error-rates is to portray UCC as the best, which doesn't necessarily feel like a fair comparison. I think we need a stronger reason to use different error rates. I'm sure there are good ones, I just don't know them.

If we do keep different error-rates, then I think it makes sense to record that data in the results files.

Misty-W: In light of this discussion, I will open a new PR to change the error rates back to the global variables.

PR #231, opened 2025-02-13
Fix relative path bug in expval benchmarking script

natestemen:
In order to run this script as part of the github actions pipeline, it should be runnable from any location.


bachase (approved):
👍

PR #244, opened 2025-02-18
Draft: Benchmark update dependency

bachase:


bachase:
I'll remove [bae3d42](https://github.com/unitaryfund/ucc/pull/244/commits/bae3d4212682753b5c6130f78bf4a0df24f5facf) before committing, but you will see that the labels for cirq are a bit confusing now given that premature run with development version.

natestemen (approved):
Good catch on the cirq pre-release. 🚢

Issue #242, opened 2025-02-18
Make comparison benchmark code for external compiler configurations clearer


willzeng:
Right now this code is a bit hidden: https://github.com/unitaryfund/ucc/blob/main/benchmarks/scripts/common.py#L89-L110

This could 
(1) be refactored to make it easier for contributors to suggest updates to the defaults that we use and 
(2) be included in the docs so that contributors can easily find what defaults we mean when we mention qiskit and pytket on the benchmark plots

bachase:
I'll focus on option 2 for as part of the next milestone, but will consider a refactored version as part of #235

bachase:
Closing this as done as part of #313

Issue #202, opened 2025-01-28
Add documentation on where UCC fits into ecosystem


natestemen:
Something to showcase where UCC is a drop in replacement for other tools, and where it isn't. A survey of adjacent tools.

suggested by @nathanshammah

bachase:
@nathanshammah for a first pass, are you thinking to compare it to the compilers in the benchmark set?

natestemen:
I'll let Nathan comment, but I think a possible starting point could be the literature review that was done at the start of the project. I believe [this document](https://docs.google.com/document/d/1kw2QkREjYA5vBCeCqzDOeilyNczal9grnillU7QCsRg/) is that review, but there may be something else. @jordandsullivan and @nathanshammah to comment.

bachase:
Any concerns with moving this to the next milestone? I'd like to get a better understanding and don't want to block 0.4.3

jordandsullivan:
We can move it to the next milestone.

bachase:
Continuing to 🤔 on this ...

The draft blog post closes with a great north star "... UCC to make quantum programming simpler, faster, and more scalable." In the current README, we have "frontend-agnostic, high performance compilation of quantum circuits.", which might understate some of the focus on simplicity and scalability, and certainly some of the benefits from being open/open source.  Combining these in the README might give a better framing. Are there any other design principles or goals the team had used outside this set (simple, fast, scalable, open)? Extensible (which could be part of open + simple)?

Separately, for the docs at this beginning stage, we can also consider a page for FAQs, such as

1. "Why should I use ucc instead of XXX, YYY?"
2. "Why does ucc build on top of existing compiler technologies?"

to address common questions around `ucc` purpose and design.

jordandsullivan:
I agree with this approach. Yes, indeed, interoperability and extensibility are primary design goals and I think it makes sense to call them out as such.

jordandsullivan:
I like the idea of an FAQ section as well, even if that goes into a separate issue.

bachase:
@jordandsullivan thinking this could be a step towards review paper?

PR #245, opened 2025-02-19
Update documentation to expand on design goals:

bachase:
Updates the README to add more context on UCC's goals and place in the ecosystem.

Adds a benchmarking section to the docs to give more details on the benchmark suite design and how users can contribute or suggest improvements.

Attempts to address both #242 and #202.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,14 @@
 [![Discord Chat](https://img.shields.io/badge/dynamic/json?color=blue&label=Discord&query=approximate_presence_count&suffix=%20online.&url=https%3A%2F%2Fdiscord.com%2Fapi%2Finvites%2FJqVGmpkP96%3Fwith_counts%3Dtrue)](http://discord.unitary.foundation)
 
 
-The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits. UCC's goal is to make quantum programming simpler, faster, and more scalable, all while building on and contributing to the open source quantum ecosystem.
+The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits. UCC's goal is to gather together the best of open source compilation to make quantum programming simpler, faster, and more scalable.
 
 By leveraging [qBraid](https://github.com/qBraid/qBraid), UCC interfaces automatically with multiple quantum computing frameworks, including [Qiskit](https://github.com/Qiskit/qiskit), [Cirq](https://github.com/quantumlib/Cirq), and [PyTKET](https://github.com/CQCL/tket) and supports programs in OpenQASM 2 and [OpenQASM 3](https://openqasm.com/). For a full list of the latest supported interfaces, just call `ucc.supported_circuit_formats`.
 
 
 **Want to know more?**
 - Check out our [documentation](https://ucc.readthedocs.io/en/latest/), which you can build locally after installation by running `make html` in `ucc/docs/source`.
-- Read the [launch announcement](TODO) to learn more on the current state of UCC, its capabilities and future direction.
+- Read the [launch announcement](https://unitary.foundation/posts/2025_ucc_launch_blog) to learn more on the current state of UCC, its capabilities and future direction.
 - For code, repo, or theory questions, especially those requiring more detailed responses, submit a [Discussion](https://github.com/unitaryfund/ucc/discussions).
 - For casual or time sensitive questions, chat with us on [Discord](http://discord.unitary.foundation).
 
--- a/README.md
+++ b/README.md
@@ -6,13 +6,14 @@
 [![Discord Chat](https://img.shields.io/badge/dynamic/json?color=blue&label=Discord&query=approximate_presence_count&suffix=%20online.&url=https%3A%2F%2Fdiscord.com%2Fapi%2Finvites%2FJqVGmpkP96%3Fwith_counts%3Dtrue)](http://discord.unitary.foundation)
 
 
-The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits.
+The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits. UCC's goal is to make quantum programming simpler, faster, and more scalable, all while building on and contributing to the open source quantum ecosystem.
 
-UCC interfaces automatically with multiple quantum computing frameworks, including [Qiskit](https://github.com/Qiskit/qiskit), [Cirq](https://github.com/quantumlib/Cirq), and [PyTKET](https://github.com/CQCL/tket) and supports programs in OpenQASM 2 and [OpenQASM 3](https://openqasm.com/). For a full list of the latest supported interfaces, just call `ucc.supported_circuit_formats`.
+By leveraging [qBraid](https://github.com/qBraid/qBraid), UCC interfaces automatically with multiple quantum computing frameworks, including [Qiskit](https://github.com/Qiskit/qiskit), [Cirq](https://github.com/quantumlib/Cirq), and [PyTKET](https://github.com/CQCL/tket) and supports programs in OpenQASM 2 and [OpenQASM 3](https://openqasm.com/). For a full list of the latest supported interfaces, just call `ucc.supported_circuit_formats`.
 
 
 **Want to know more?**
 - Check out our [documentation](https://ucc.readthedocs.io/en/latest/), which you can build locally after installation by running `make html` in `ucc/docs/source`.
+- Read the [launch announcement](TODO) to learn more on the current state of UCC, its capabilities and future direction.
 - For code, repo, or theory questions, especially those requiring more detailed responses, submit a [Discussion](https://github.com/unitaryfund/ucc/discussions).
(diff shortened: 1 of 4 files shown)


docs/source/dev.rst
```diff
@@ -18,12 +18,12 @@ To release a new version of ``ucc`` on GitHub, follow the steps below.
     - Update the ``CHANGELOG.rst`` file with all new changes, improvements, and bug fixes since the previous release.
 
 3. **Commit Changes:**
-    - Commit the changes to `VERSION.txt` and `CHANGELOG.rst` and open a PR to get the changes reviewed.
+    - Commit the changes to ``pyproject.toml`` and `CHANGELOG.rst` and open a PR to get the changes reviewed.
```
bachase, line 21: This was a miss from earlier when we migrated to poetry

bachase:
Note that I enabled https://docs.readthedocs.com/platform/stable/guides/pull-requests.html so its easier to preview the docs changes.

README.md
```diff
@@ -6,13 +6,14 @@
 [![Discord Chat](https://img.shields.io/badge/dynamic/json?color=blue&label=Discord&query=approximate_presence_count&suffix=%20online.&url=https%3A%2F%2Fdiscord.com%2Fapi%2Finvites%2FJqVGmpkP96%3Fwith_counts%3Dtrue)](http://discord.unitary.foundation)
 
 
-The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits.
+The **Unitary Compiler Collection (UCC)** is a Python library for frontend-agnostic, high performance compilation of quantum circuits. UCC's goal is to make quantum programming simpler, faster, and more scalable, all while building on and contributing to the open source quantum ecosystem.
```
jordandsullivan, line 9: In my opinion this new line is a bit too generic and doesn't add much new information. Let's be a little bit more specific about making it easier to develop the quantum compilation stack.

docs/source/benchmarking.rst
jordandsullivan, line 1: This is really fantastic @bachase . Great way to lay out our benchmarking philosophy and guide new users through it.

docs/source/benchmarking.rst
```diff
@@ -0,0 +1,69 @@
...
+via ``apt-get install parallel`` on Ubuntu or ``brew install parallel`` on MacOS.
+
+Contributing to benchmarks
+--------------------------
+
+We welcome contributions to improve UCC’s benchmarking suite! This could include
+
+- **Adding New Benchmark Circuits** to cover an additional real-world use case.
+- **Improving Benchmark Metrics** to add additional metrics of interest to compare compiler performance.
+- **Optimizing Compiler Configuraitons** to improve the default configuration of compilers in the benchmark.
```
jordandsullivan, line 65: Typo: Configurations

jordandsullivan: Let's add links here to the code which would specify what we mean by compiler configurations. It might not be clear. For instance, that we mean for Qisket you can just say `compile(optimization level=3)`  vs. needing to Define every transpiler pass in Cirq for instance.

jordandsullivan (approved):
A few comments, but mainly this looks good to me, feels right on the positioning.

docs/source/benchmarking.rst
```diff
@@ -0,0 +1,69 @@
...
+     - Many of the included circuits are from the `benchpress <https://github.com/Qiskit/benchpress>`_ benchmarking suite.
+
+2. **Compilers**: UCC is benchmarked comparing its default passes (defined in :class:`ucc.transpilers.ucc_defaults.UCCDefault1`) against other similar defaults for leading compilers, including `Qiskit <https://github.com/Qiskit/qiskit>`_, `Cirq <https://github.com/quantumlib/Cirq>`_, and `PyTKET <https://github.com/CQCL/tket>`_.
+      - The current compiler settings are defined `here <https://github.com/unitaryfund/ucc/blob/fde89f6e25f3adc2b47313e3e1c7cc0b5b2e1a18/benchmarks/scripts/common.py#L90-L207>`_ and we welcome contributions to these defaults as suggested below.
+
+3. **Metrics**: We track several metrics to evaluate compiler performance:
+    - *Compiled Gatecount Ratio*: Measures the ratio of 2-qubit gates in the compiled versus raw circuit.
+    - *Compilation Time*: Tracks the time taken to compile a circuit.
+    - *Observable under noise*: Measures the fidelity of the compiled circuit under noise models, using an observable relevant for that circuit.
```
natestemen, line 34:
```suggestion
    - *Observable under noise*: Measures the fidelity of the compiled circuit under a depolarizing noise model, using an observable relevant for that circuit.
```

Currently only one noise model in use. Should we be explicit here or is it okay to be generic?

docs/source/benchmarking.rst
```diff
@@ -0,0 +1,69 @@
...
+
+4. **Reproducibility**: In order to ensure the reliability of our benchmarks, we follow these practices:
+    - Benchmark scripts and results are available in the `UCC repository <https://github.com/unitaryfund/ucc/tree/main/benchmarks>`_.
+    - Results are stored in version-controlled reports to track improvements over time.
+    - Official results are run using a dedicated GH runner for consistent comparison.
```
natestemen, line 39:
```suggestion
    - Official results are run using a dedicated machine for consistent comparison.
```

Just thinking about readers not intimately familiar with GitHub. If you want to keep reference to the github runner, lets use GitHub instead of GH.

docs/source/benchmarking.rst
```diff
@@ -0,0 +1,69 @@
...
+To run the benchmarks locally, first follow the steps in the
+:doc:`contributing guide <contributing>` to setup your dev environment. Then, run
+the benchmark script via:
+
+.. code-block:: sh
+
+   poetry run ./benchmarks/scripts/run_benchmarks.sh 8
+
+This will run the benchmarks in parallel on 8 processes, and store the results to
+``benchmarks/results/`` as CSV files.
```
natestemen, line 53: Why 8? I had the impression that the script automatically figured out how parallel it can run the benchmarks.

natestemen (approved):
This is great documentation. A few changes/questions, but 👍🏻 from me!

docs/source/benchmarking.rst
bachase: How about splitting the difference and going with "...the compiled circuit under noise"

I think being generic is fine, as we are likely to change this over time.

docs/source/benchmarking.rst
bachase: The current script defaults to 4 https://github.com/unitaryfund/ucc/blob/5b992eb45d42eaf79aeb949d91d29fa771147d8b/benchmarks/scripts/run_benchmarks.sh#L29

I must've grabbed 8 from the benchmark GH action definition. I can leave it out for now to avoid a misleading number, and then we can choose to update the script to query the number of cores to run on as a separate followup.

bachase: Actually, deciding to just instruct the user to specify the desired parallelism. (will push this suggestion shortly)

README.md
bachase: Fair point. Do you think its worth having this line then? Or just rely on the information further down? I think the example code better gives that specificity. 

But at the same time, would be good to highlight that as a key aspect (and the open source aspect) from the get go.

🤔

docs/source/benchmarking.rst
natestemen: Sounds good.

docs/source/benchmarking.rst
bachase: I updated this with a few more details in the latest commit.

README.md
jordandsullivan: I think it would be good to have a bit of framing specifically outlining how we are aiming to provide a unified interface for user and repository for compiler developers to contribute their passes.

jordandsullivan: Something like "UCC goal is to provide a multi-framework interface and open source repository for quantum compiler users and researchers." 
It could still use some word smithing, but I do think these goals are important to highlight.
Currently lots of compiler development goes on independently and is not compatible across different frameworks. We want to encourage the research community to start contributing their newly developed compiler passes to UCC and in the longer term to joining us in defining standards for intermediate representations and other layers of the quantum stack.

Misty-W: +1 for Jordan's comments.

Misty-W (approved):
Agree with Jordan and Nate's suggestions, once resolved I'd say this is ready to go. Great improvements to the narrative and to see this evolve as a "living" description of UCC!

README.md
willzeng: I agree that highlighting the key aspects are good to do upfront here. How about something like:

"UCC's goal is to gather together the best of open source compilation to make quantum programming simpler, faster, and more scalable."

willzeng (approved):
Changes broadly LGTM!

Small comments that can be taken or left.

README.md
bachase: I went with @willzeng suggestion for now to keep it simple but still capture these themes.

Perhaps as an alternative, we can leave #202 open and have a future update that adds more of that motivation/context to the docs build, where we can expand on those themes.

bachase:
@jordandsullivan in this latest push, I used the anticipated URL (https://unitary.foundation/posts/2025_ucc_launch_blog) for the launch blog. That will 404 for now, but I think thats ok for few days. 

Unless any concerns, I'll plan to merge this later today.

PR #247, opened 2025-02-19
Ensure benchmarking runs don't add gitignored files

bachase:
As discussed in https://github.com/unitaryfund/ucc/pull/223#discussion_r1960590003, this changes the benchmark runs to not force commit all changes, which was avoiding filters in gitignore.


natestemen (approved):

jordandsullivan (approved):
Thanks for fixing!

Issue #222, opened 2025-02-11
Modify noise strength for circuit expecation values


natestemen:
Set the noise strength so Qiskit compiler + simulation gets within 10% of ideal value. This value will need to be found on a per-circuit basis.

natestemen:
closing this in favor of https://github.com/unitaryfoundation/ucc/issues/299

PR #248, opened 2025-02-20
Change QV error rates back to global variables

Misty-W:
Per [comments](https://github.com/unitaryfund/ucc/pull/223#discussion_r1960319404) in #223, we need a stronger justification for varying error rates. We may wish to do so later to solve #222 but it needs more investigation.

This PR is to use the global error rates consistently for noisy simulation in `expval_benchmark.py`.


bachase (approved):
LGTM

natestemen (approved):

Issue #125, opened 2024-12-06
Add dependabot


jordandsullivan:
**Is your feature request related to a problem? Please describe.**
We always want to remain up to date with the latest versions of other compilers we benchmark against, and especially packages we depend on deeply (like Qiskit at this point in time).

**Describe the solution you'd like**
Mostly likely using Dependabot is the way to go, as [Mitiq](https://github.com/unitaryfund/mitiq) already uses.

**Describe alternatives you've considered**
I suppose the alternative is checking this by hand or relying on our benchmarking pipeline to break, but automating it seems better.

**Additional context**

PR #250, opened 2025-02-20
Upgrade dependencies

### LINKED ISSUE #233: Improve performance reading qasm
**Is your feature request related to a problem? Please describe.**

When running some of the benchmarking code, I observed that the wallclock times of the runs are significantly longer than the measured compilation time. On profiling, it turns out this is due to how qbraid is determining the version of the qasm file when doing the initial translation between circuit representation. 

This issue is upstream in pyqasm https://github.com/qBraid/pyqasm/issues/134. If we want to prioritize, we can contribute the fix there. Once in a new pyqasm/qbraid release, we would want to upgrade to that version.

bachase:
#125 will automate this, but for now, upgrading to the latest version of pyqasm and qbraid to benefit from improvements in parsing speed during benchmark runs.

Fixes #233

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1057,22 +1057,22 @@ files = [
 
 [[package]]
 name = "openqasm3"
-version = "1.0.0"
+version = "1.0.1"
 description = "Reference OpenQASM AST in Python"
 optional = false
 python-versions = "*"
 groups = ["main"]
 files = [
-    {file = "openqasm3-1.0.0-py3-none-any.whl", hash = "sha256:d4371737b4a49b0d56248ed3d30766a94000bccfb43303ec9c7ead351a1b6cc3"},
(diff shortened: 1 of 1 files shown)


jordandsullivan (approved):
Thanks, lgtm!

jordandsullivan:
Taking a look at the latest benchmarks, it looks like PyTKET suddenly slower with these updates. There wasn't a change to the PyTKET version though, so I'm not sure what could be causing this.

bachase:
> Taking a look at the latest benchmarks, it looks like PyTKET suddenly slower with these updates. There wasn't a change to the PyTKET version though, so I'm not sure what could be causing this. 

Yes, but also looks like some drift for others over the last few runs as well? I suggest we consider using the pinned versions in the lock file to at least remove that degree of freedom in case some of the common dependencies are changing silently? (We can enable dependabot to get them auto-upgrading in a more controlled manner).

I'm less familiar with how consistent the runner instance is, and how well we are isolating the individual runs

bachase:
I wonder if the noise simulations have some internal parallelization too. I can make a new ticket to investigate this tomorrow

Issue #251, opened 2025-02-21
Investigate variance in benchmarking results


bachase:
Taking a look at the latest benchmarks, it looks like PyTKET suddenly slower with these updates. There wasn't a change to the PyTKET version though, so I'm not sure what could be causing this. 
![17400910229213713367557020631448](https://github.com/user-attachments/assets/7f8fb5ba-9132-4152-a0fa-009555177caa)

_Originally posted by @jordandsullivan in https://github.com/unitaryfund/ucc/issues/250#issuecomment-2672853194_

bachase:
Haven't found a clear issue yet, but noticed a few things that could cause uncontrolled sources of variance

1. Multiple benchmark actions could be in parallel (look into [controlling job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs))
2. The docker container built for each run pulls the latest compatible libraries. Even if the main compiler dependency versions are unchanged, its possible the transitive dependencies could impact the performance? Say if qbraid changed how it translates circuits between compiler libraries, that would change the outputs?
3. Qiskit-Aer simulation has some support for parallelization (see options [here](https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html#qiskit_aer.AerSimulator)). I don't believe these are getting enabled by default or in our configurations, but still understanding what gets parallelized at the python level versus what might be happening in the C++

bachase:
I was able to recreate a bit of this locally by focusing on the large change in `pytket` running time recently. Locally, I found that varying the parallelism of the script changes this result. Looking into the `Qiskit-Aer` options as previously mentioned, it does look like if OpenMP is enabled, by default the simulation tries to use all available threads (see [here](https://github.com/Qiskit/qiskit-aer/blob/fa271c7cffc0b05e6628809139e0fb6957241947/src/controllers/aer_controller.hpp#L206-L216)). For these benchmarks, I believe this only matters for the statevector simulation of HOV which repeats several shots.

To test this on the GIthub Runner, I made a change to set the max threads to 1 when constructing the AERSimulator instances (https://github.com/unitaryfund/ucc/tree/bc-test-AER-1-thread). Running the benchmarks shows this reverts back to the prior timing

Baseline (current master branch)
[...]

PR #253, opened 2025-02-24
Improve reproducibility and consistency of benchmarking workflow

bachase:
These changes will
1. Reduce the parallelism in the benchmarking runs to minimize contention that might otherwise impact the results. This includes restricting Qiskit-AER simulations to a maximum of 1 parallel thread and run expected value benchmarks separate in parallel separate from timing benchmarks in parallel. ~~and forcing just 1 parallel job in the benchmark script itself~~.

2. Use the locked version of dependencies in the benchmarking runs to avoid an uncontrolled change in a dependent library or transitive dependency.

3. Ensure only one instance of the benchmark github workflow runs at a time by specifying a concurrency group.

This partially addresses #251

### DIFF
--- a/.github/workflows/ucc-benchmarks.yml
+++ b/.github/workflows/ucc-benchmarks.yml
@@ -12,6 +12,10 @@ on:
         required: false
         default: 'main'
 
+# Ensure only one instance of the workflow is running at a time
+concurrency:
+  group: ${{ github.workflow }}
+
 jobs:
   run-benchmarks:
     runs-on: ucc-benchmarks-8-core-U22.04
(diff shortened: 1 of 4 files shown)


natestemen (commented):
Makes sense to me. Do you know by how much time this increases the runtime of `run_benchmarks.sh`?

bachase:
> Makes sense to me. Do you know by how much time this increases the runtime of `run_benchmarks.sh`?

Looking at https://github.com/unitaryfund/ucc/actions/workflows/ucc-benchmarks.yml?query=branch%3Abc-test-AER-1-thread, from about 10 minutes to about 30 minutes. 

Per some more recent discussion in #251, we could 

1. Retain the parallelism for the expected value benchmark runs as those are not measuring timing
2. Retain parallelism for all benchmark runs, but just schedule the expected value ones separate from the timing ones.

Option 2 would keep the overall runtime fastest, but option 1 is more conservative until we get a stronger handle on how the parallel jobs are getting scheduled on the runner. But given the historical benchmark runs, this might be overly conservative as things were fairly steady for a while.

natestemen (approved):

PR #254, opened 2025-02-24
Change plot resolution to per release in plot over time scripts

### LINKED ISSUE #200: Update benchmark plot resolution to monthly
Perhaps updating this to instead plot by month will scale better over time:

Image

Otherwise will get quite long quickly

Misty-W:
Fixes #200.

Change plot resolution to per compiler release in plot over time scripts, so the plots will look like the following:
image
image

Note: I pasted (instead of pushed) these plots because they'd have to be removed before merge anyway.

### DIFF
--- a/benchmarks/scripts/plot_avg_benchmarks_over_time.py
+++ b/benchmarks/scripts/plot_avg_benchmarks_over_time.py
@@ -179,7 +179,9 @@ for compiler in unique_compilers:
 
 # Add annotations for version changes
 for date in filtered_avg_compile_time["date"].unique():
-    date_data = filtered_avg_compile_time[filtered_avg_compile_time["date"] == date]
+    date_data = filtered_avg_compile_time[
+        filtered_avg_compile_time["date"] == date
+    ]
     # Sort date_data in order of compiled_ratio
     date_data = date_data.sort_values("compile_time")
     for _, row in date_data.iterrows():
--- a/benchmarks/scripts/plot_avg_benchmarks_over_time.py
+++ b/benchmarks/scripts/plot_avg_benchmarks_over_time.py
@@ -110,48 +110,43 @@ for compiler in unique_compilers:
         color=color_map[compiler],
     )
 
+# Keep track of last compiler version seen over time to track version changes
+last_version_seen = {compiler: None for compiler in unique_compilers}
 
 previous_bboxes = []
+# Now iterate over each compiler entry and annotate if it's a new version, if so label it
+for date in avg_compiled_ratio["date"].unique():
+    date_data = avg_compiled_ratio[avg_compiled_ratio["date"] == date]
+    # Sort date_data in order of compiled_ratio
+    date_data = date_data.sort_values("compiled_ratio")
+    # Now iterate over each compiler entry and annotate if it's a new version
+    for _, row in date_data.iterrows():
+        # Get the version for this date
+        current_version = row["compiler_version"]
+        compiler = row["compiler"]
+        compiled_ratio = row["compiled_ratio"]
 
-# Find the latest version for each package
-latest_versions = avg_compiled_ratio.groupby("compiler")[
-    "compiler_version"
-].max()
-# Filter to keep only rows with the latest version
-avg_compiled_ratio_latest_versions = avg_compiled_ratio[
-    avg_compiled_ratio["compiler_version"].isin(latest_versions)
-]
-#  Find the first appearance date of each latest version
-first_appearance_dates = (
-    avg_compiled_ratio_latest_versions.groupby(
-        ["compiler", "compiler_version", "compiled_ratio"]
-    )["date"]
-    .min()
-    .reset_index()
-)
-
-
-for _, row in first_appearance_dates.iterrows():
-    # Get the version for this date
-    current_version = row["compiler_version"]
-    compiler = row["compiler"]
-    compiled_ratio = row["compiled_ratio"]
-
-    text = f"{compiler}={current_version}"
-    xy = (row["date"], compiled_ratio)
-    color = color_map[compiler]
-    # Add the annotation and adjust for overlap
-    annotate_and_adjust(
-        ax=ax[0],
-        text=text,
-        xy=xy,
-        color=color,
-        previous_bboxes=previous_bboxes,
-        offset=(0, 20),  # Initial offset
-        increment=2,  # Vertical adjustment step
-        max_attempts=15,
-    )
-    # plt.pause(0.1)
+        # Check if the version has changed
+        if current_version != last_version_seen[compiler]:
+            text = f"{current_version}"
+            xy = (row["date"], compiled_ratio)
+            color = color_map[compiler]
+            renderer = fig.canvas.get_renderer()
+            previous_bboxes.append(ax[0].get_tightbbox(renderer))
+            # Add the annotation and adjust for overlap
+            annotate_and_adjust(
+                ax=ax[0],
+                text=text,
+                xy=xy,
+                color=color,
+                previous_bboxes=previous_bboxes,
+                offset=(0, 20),  # Initial offset
+                increment=2,  # Vertical adjustment step
+                max_attempts=15,
+            )
+            # plt.pause(0.1)
+            # Update the last seen version for this compiler
+            last_version_seen[compiler] = current_version
 
 # Set y axis range to be slightly larger than data range
 adjust_axes_to_fit_labels(ax[0], y_scale=[1.1, 1.3])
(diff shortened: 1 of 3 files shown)


bachase:
For the absolute error plot, does each dot still represent a new version of that software package? If so, would we want it to have the label as well?

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -62,17 +60,27 @@ def extract_compiler_versions(header):
...
+    df_versions = df_dates[df_dates["compiler_version"] == version]
+    earliest_date = sorted(df_versions["date"].unique())[0]
+    dates.append(earliest_date)
 
 fig, ax = plt.subplots(2, 1, figsize=(8, 8), sharex=False, dpi=150)
 # Rotate x labels on axes 0
 plt.setp(ax[0].xaxis.get_majorticklabels(), rotation=45)
 
 
+filtered_avg_compiled_ratio = avg_compiled_ratio[
```
bachase, line 75: How are we thinking about which dates are included in the earlier average and this filter?

`avg_compiled_ratio` is calculated earlier as:
```python
avg_compiled_ratio = (
    df_dates.groupby(["compiler", "date", "compiler_version"])[
        "compiled_ratio"
    ]
    .mean()
    .reset_index()
    .sort_values("date")
)
```

If we happened to have run multiple benchmarks for that compiler on that date, where say some were for version 1.2.0 and then some for version 1.3.0, would this filter be showing results that average over both versions?

Would we also need to filter to only include/average results from that date where the latest compiler version was used?

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -62,17 +60,27 @@ def extract_compiler_versions(header):
 color_map = {
     compiler: colormap(i) for i, compiler in enumerate(unique_compilers)
 }
+unique_versions = sorted(df_dates["compiler_version"].unique())
+dates = []
+for version in unique_versions:
+    df_versions = df_dates[df_dates["compiler_version"] == version]
```
bachase, line 66: Is this `compiler_version` only the semver string? If so, is it possible that this might mix the same version of different packages? For example `cirq 1.2.0` and `qiskit 1.2.0`? If so, I think that might end up taking the earliest date across the union of those two.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -151,8 +166,8 @@ def extract_compiler_versions(header):
     )
 
 # Add annotations for version changes
-for date in avg_compile_time["date"].unique():
-    date_data = avg_compile_time[avg_compile_time["date"] == date]
+for date in filtered_avg_compile_time["date"].unique():
+    date_data = filtered_avg_compile_time[avg_compile_time["date"] == date]
```
bachase, line 170: I think this is mixing dataframes, consider:
```suggestion
    date_data = filtered_avg_compile_time[filtered_avg_compile_time["date"] == date]
```

bachase (changes requested):
The comments on the logic for selecting and filtering which benchmark results end up included in the average performance on earliest version release are my main open questions.

**edit** To be clear, its only the date and version selection logic that I wanted to understand as part of this review. The discussion below is more food for thought for the future.

But stepping back, I think this may still be fragile or at least challenging given how the benchmark results are stored by the date of the run currently. What if some days have more benchmark runs than others? If so, will that be a fair comparison if some days average over many runs and others only a single run? 

As I (poorly) started to sketch out in #251, I think we should consider switching to associate benchmark results with the git hash (and corresponding git commit time) to ensure a singular mapping from set of compiler versions (as reflected in the lock file in the commit) to benchmark results. Then we can more cleanly pick out results when versions change and plot accordingly.  

That is clearly a larger scope than this change, but I suspect getting the plot we want will be hard using the current way we run and store benchmarks. Working around it might be possible, but the code will get more complicated and arguably worth testing to be sure all the cases are handled.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
jordandsullivan: Because this groups by the date, compiler name, and compiler version, it would not average two different versions together. On the plot I think it would just show them as different datapoints in the same color on the same date and label them correspondingly.

benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -33,9 +39,18 @@
 # Convert the 'date' column to datetime
 df_all["date"] = pd.to_datetime(df_all["date"])
 
+unique_versions = sorted(df_all["compiler_version"].unique())
```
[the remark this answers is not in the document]
jordandsullivan: Good catch, these are just the numbers:
`Here are unique_versions ['0.1.0', '0.1.1', '0.2.0', '0.4.0', '0.4.2', '1.2.0', '1.3.0', '1.3.1', '1.3.2', '1.3.3', '1.35.0', '1.36.0', '1.37.0', '1.39.0', '1.4.0', '1.4.1', '1.40.0', '1.5.0.']`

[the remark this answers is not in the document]
Misty-W: I'm aware these are only the numbers, but the actual filtering is done on the date each new version appears in the dataset:
```
dates = []
for version in unique_versions:
    df_versions = df_dates[df_dates["compiler_version"] == version]
    earliest_date = sorted(df_versions["date"].unique())[0]
    dates.append(earliest_date)
filtered_avg_compiled_ratio = avg_compiled_ratio[
    avg_compiled_ratio["date"].isin(dates)
]
```

benchmarks/scripts/plot_avg_benchmarks_over_time.py
Misty-W: That's correct @jordandsullivan .

bachase: Got it. Thank you both.

If we had multiple datapoints, would we want to only plot the "latest version" dot for that date? Just wondering if it might get cluttered otherwise.

benchmarks/scripts/plot_expval_benchmarks_over_time.py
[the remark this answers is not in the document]
bachase: Here's a toy example to consider (ignoring the actual data columns)
```csv
date, compiler, compiler_version
2025-01-01,cirq,1.2.0
2025-01-02,qiskit,1.2.0
2025-01-03,cirq,1.3.0
```

I think this might select only `2025-01-1` and `2025-01-03` as earliest dates, as the cirq version 1.2.0 would be earlier than the qiskit 1.2.0. (In this parallel universe where their versions are close and overlapping)

Misty-W:
I changed the dataframe `groupby` method to group both by compiler and version number, so now it only plots points with dates corresponding to a new compiler version used in the benchmarks.

I've also changed the version labels so that they only show the latest version, to avoid crowding the plot with labels.

image

jordandsullivan:
Suggest keeping the versions, but removing the name of the compiler from the label, since it is already present in the legend.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -116,10 +128,11 @@ def extract_compiler_versions(header):
 
         # Check if the version has changed
         if current_version != last_version_seen[compiler]:
-            text = f"{compiler}={current_version}"
+            text = f"{current_version}"
             xy = (row["date"], compiled_ratio)
             color = color_map[compiler]
-
+            renderer = fig.canvas.get_renderer()
+            previous_bboxes.append(ax[0].get_tightbbox(renderer))
```
bachase, line 135: Can you share more on what this line is doing (adding the current tightbbox)? Is there a need to do it in the average compile time plot too?

benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -62,7 +80,54 @@
...
+    summary["compiler_version"].isin(latest_versions)
+]
+#  Find the first appearance date of each latest version
+first_appearance_dates = (
+    avg_absolute_error_latest_versions.groupby(
+        ["compiler", "compiler_version", "avg_absolute_error"]
+    )["date"]
+    .min()
+    .reset_index()
+)
```
bachase, line 99: Are these lines still needed?

bachase (commented):
Generally 👍. 

I did notice when running locally for me, there's still a good bit of overlap with all the versions.

Have we tried an off-the-shelf tool like https://adjusttext.readthedocs.io/en/latest/index.html to manage this versus the `annotate_and_adjust` approach?

image

jordandsullivan:
> Have we tried an off-the-shelf tool like https://adjusttext.readthedocs.io/en/latest/index.html to manage this versus the `annotate_and_adjust` approach?
> 
There's an open issue for this: https://github.com/unitaryfund/ucc/issues/271

benchmarks/scripts/plot_avg_benchmarks_over_time.py
Misty-W: It was a failed experiment to get the labels to render more nicely. I've removed it.

benchmarks/scripts/plot_expval_benchmarks_over_time.py
Misty-W: Good catch, no longer needed and removed.

bachase:
> should we filter those out also?

I'm supportive, especially as it was unintended/not reflecting a meaningful change.

bachase (approved):
LGTM and support removing (or not removing) those unintended versions.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -60,6 +72,14 @@ def extract_compiler_versions(header):
     .reset_index()
     .sort_values("date")
 )
+# Sort the dataframe by package, date, and version
+avg_compiled_ratio = avg_compiled_ratio.sort_values(
```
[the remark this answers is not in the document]
Misty-W: Thanks for flagging this. The standard library does have `parse_version`,  although it is deprecated. For this PR, I decided it best to use `parse_version`, so at least we're sorting correctly, and then do a follow up PR to replace the deprecated method with the method from `packaging`.

Misty-W:
After filtering out the data from March 5, we have more sensibly labeled plots:

image

Issue #246, opened 2025-02-19
Clarify content of transpiler passes in documentation


nathanshammah:
**Is your feature request related to a problem? Please describe.**
Currently the readme and documentation (user guide, API doc) are quite minimal on the explanation of the transpiler passes and defaults present in `ucc`. 

**Describe the solution you'd like**
The documentation can be improved with a summary of existing transpiler passes. We want to clarify:
- If the transpiler pass comes from another library, as currently for Qiskit ones, why it was chosen
- If the pass was customized with input variables, e.g., 
- If the pass was modified and why, e.g., VF2
- Choices made in ucc [defaults](https://github.com/unitaryfund/ucc/blob/fde89f6e25f3adc2b47313e3e1c7cc0b5b2e1a18/ucc/transpilers/ucc_defaults.py)

**Additional context**
Useful comment in the Discussions https://github.com/unitaryfund/ucc/discussions/239.

jordandsullivan:
New section in User Guide on Transpile Passes:
- Clarifies what each pass does that is _different_ from default Qiskit

bachase:
Perhaps a new consideration in light of #256 is to supplement https://github.com/unitaryfund/ucc/blob/main/docs/source/user_guide.rst#key-modules with more details on why the specific set of passes was chosen for the `UCCDefault1` set?

jordandsullivan:
Yes, that makes sense as a new direction for this part of the documentation. As I recall, these passes were chosen partially by doing some empirical tests of which Qiskit passes seemed to provide the most bang for your buck (i.e. where is most of the gate reduction happening). Is that accurate @sonikaj?

sonikaj:
Jordan, that's correct and also it comes from consideration of what would be a logical sequence of operations to optimize a general quantum circuit about which not much other information is known. If there is some other information is available at compile time, for example, that it is a chemistry circuit, or it's implementing a Grover's type algorithm, a different sequence of passes, custom to the particular application area, can be designed.

PR #256, opened 2025-02-25
Remove custom transpilation passes

bachase:
As part of #246, the team benchmarked the impact of the customized transpilation passes compared with the version of the equivalent passes in the latest version of Qiskit. The results show the customizations are not currently providing additional gate savings or compilation time improvements compared with the native Qiskit passes. 

This PR removes these customized passes and replaces them with the Qiskit equivalents in the default set in `UCCDefault1`.


jordandsullivan:
I just manually triggered a benchmark run to check performance, and it looks good: same gate reduction and faster runtime. Although we are no longer focused on the narrative of being "the best compiler," this change actually brings us to be the top performing compiler in terms of speed and compiled ratio for our benchmarks: https://github.com/unitaryfund/ucc/blob/6c2c00ca9a7680275decda2dce30878241accd18/benchmarks/avg_compiler_benchmarks_over_time.png

Also, perhaps we should consider adding the Run benchmarks step as an automatic step in PRs -- only reason I didn't was that I didn't want them to run every time code was pushed to an open PR, plus it can create annoying merge conflicts when replacing the .pngs.

jordandsullivan (approved):
Thanks for doing this investigation and finding that we can get this speedup by using the latest Qiskit versions of our UCC pass set.

bachase:
Thanks! 

I agree on having these run as part of PRs. I think I've got a suggestion on how to do it without causing the conflicts, but need to write it down to validate in #235. Hoping to get my thoughts in markdown this week/next at the latest.

PR #260, opened 2025-02-26
Change QAOA observable to problem Hamiltonian

### LINKED ISSUE #255: Custom observable for QAOA
Change benchmarks/scripts/expval_benchmark.py so that if the compiled circuit is a QAOA circuit, it applies the problem Hamiltonian from The same process is repeated with the uncompiled circuit but on a noiseless simulator for comparison.

Misty-W:
Fixes #255. 

A natural choice of observable for the QAOA circuits is the problem Hamiltonian, instead of the default "ZZZZ..." observable. 

In https://arxiv.org/pdf/2009.01095, we see that the problem Hamiltonian for a weighted graph with binary encoding is:
```math
H_P = \sum_{i,j \in E} w_{i,j}H_{i,j}
```
where the matrix $H_{i,j}$ is a diagonal matrix modeling the interaction between vertices $i$ and $j$, and $w_{i,j}$ is the weight of the edge between vertices i and j as well as the resulting unitary evolution.

For $k = 2$, which corresponds to one qubit per vertex of the graph, the local Hamiltonian is

```math
H_{ij} = diag([1, -1, -1, 1])
```
which can be written
```math
H_{ij} =  \sigma_z(i) \sigma_z(j)
```


natestemen (commented):
Can you help me understand how this observable is chosen?

Starting from the circuit we are running: [qaoa_barabasi_albert](https://github.com/unitaryfund/ucc/blob/1c1b6f172569646ee47095051f456da6a4cdefa4/benchmarks/qasm_circuits/qasm2/benchpress/qaoa_barabasi_albert_N10_3reps_basis_rz_rx_ry_cx.qasm) which comes from [benchpress](https://github.com/Qiskit/benchpress/tree/078ad4de2347a711d477245c77405412ecfb85f6/benchpress/qasm/qaoa). I don't know much about this circuit, but I assume it is trying to solve some problem related to the [Barabási-Albert model](https://en.wikipedia.org/wiki/Barab%C3%A1si%E2%80%93Albert_model). Naively, I wouldn't expect this QAOA circuit to have the same Hamiltonian as a QAOA circuit solving a MAXCUT problem. My knowledge of QAOA is quite limited, so I may be missing something.

It's unfortunate that there isn't more documentation on this circuit in benchpress.

Misty-W:
> Can you help me understand how this observable is chosen?

> It's unfortunate that there isn't more documentation on this circuit in benchpress.

Since I was unable to find the problem Hamiltonian or any other observable used in Benchpress, I looked in the literature for QAOA approaches to solving Barabasi-Albert weighted graphs and found the paper I linked in this PR.
Although the authors use MAXCUT to illustrate the binary encoding and "one-hot" encoding methods, the authors mention that they applied both methods to solving a Barabasi-Albert weighted graph and report the results.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -235,8 +235,37 @@ def simulate_expvals(
 
     else:
         density_matrix = simulate_density_matrix(compiled_circuit)
-        obs_str = "Z" * compiled.num_qubits
```
natestemen, line 238: The variable `obs_str` still needs to be set in order to be recorded in the results.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -235,8 +235,37 @@ def simulate_expvals(
...
-        observable = Operator.from_label(obs_str)
+        if circuit_name == "qaoa":
+            observable = Operator(
```
natestemen, line 239: This observable only represents an operator acting on 4 qubits, but the circuit we are running is 10 qubits. Do we just assume the operator acts as identity on the other 6 qubits?

natestemen (changes requested):
Ah okay, thanks for the explanation, that helps me understand much more (and more than the reading I was doing). I wonder if we should reach out to the benchpress team to get some info on how the qaoa circuits are generated and ask what Hamiltonian they are using.

I have at one blocking comment, and one question.

benchmarks/scripts/expval_benchmark.py
Misty-W: Good catch. I wonder why it didn't raise an error when I ran it. Let me check again...

benchmarks/scripts/expval_benchmark.py
Misty-W: No, the input vector to `np.diag()` is length 20, creating a 20x20 matrix, corresponding to an operator acting on 10 qubits.

Misty-W: Yeah, I was trying to extend one of the examples to 10 qubits, but I think it's better to just use the formula they gave instead.

benchmarks/scripts/expval_benchmark.py
Misty-W: Both issues are fixed- 
1. I set `obs_str` to `"H_p"` (problem Hamiltonian) because the sum of Pauli strings is way too long.
2. The missing `obs_str` didn't throw an error because the 'if' statement should have read `if circuit_name == "qaoa_barabasi_albert"` not if circuit_name == "qaoa"`

benchmarks/scripts/expval_benchmark.py
Misty-W: I update the description of the PR with this info as well, but for a 10 qubit circuit solving a Barabasi-Albert graph, the authors use 1 qubit per vertex, which corresponds to k=2, i.e. a 2-local Hamiltonian, which looks like 
For $k = 2$, which corresponds to one qubit per vertex of the graph, the local Hamiltonian is

```math
H_{ij} = diag([1, -1, -1, 1]) =  \sigma_z(i) \sigma_z(j)
```

Misty-W: And we can get the observable $H_p$ from 
```math
H_P = \sum_{i,j \in E} w_{i,j}H_{i,j}
```
I used Qiskit's `SparsePauliOp` to combine the Pauli strings $H_{ij}$ into $H_p$

benchmarks/scripts/expval_benchmark.py
```diff
@@ -207,6 +207,50 @@ def estimate_heavy_output_prob(
...
+        (4, 5, 1.134),
+        (4, 6, 6.865),
+        (5, 6, 8.184),
+        (5, 7, 9.459),
+        (6, 7, 2.268),
+        (6, 8, 8.197),
+        (6, 9, 1.212),
+        (7, 9, 4.265),
+        (8, 9, 1.690),
+    ]
```
natestemen, line 239: Where do these weights come from? Should we add that as a comment in the code?

benchmarks/scripts/expval_benchmark.py
```diff
@@ -207,6 +207,50 @@ def estimate_heavy_output_prob(
...
+        # Start with identity string
+        pauli_string = ["I"] * num_qubits
+        # Place Z operators on the chosen qubits
+        pauli_string[i] = "Z"
+        pauli_string[j] = "Z"
+        # Convert to PauliSumOp
+        pauli_strings.append("".join(pauli_string))
+    observable = SparsePauliOp(
+        pauli_strings, coeffs=list(zip(*weighted_edges))[2]
+    )
```
natestemen, line 250:
```suggestion
    coeffs = [weight for _, _, weight in weighted_edges]

    observable = SparsePauliOp(pauli_strings, coeffs)
```

WDYT? Reads a little easier for me.

natestemen (approved):
Nice! I was initially a little worried about using [`SparsePauliOp`](https://github.com/Qiskit/qiskit/blob/stable/1.4/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py#L50-L1193) instead of an instance of [`Operator`](https://github.com/Qiskit/qiskit/blob/091228cf15c08091cb8bf832c362984f81d6a2b0/qiskit/quantum_info/operators/operator.py#L42) (since that what [`density_matrix.expectation_value`](https://github.com/Qiskit/qiskit/blob/ff15ce40479324449ec8ad8205c90d2a7b02fb11/qiskit/quantum_info/states/densitymatrix.py#L400-L421) expects), but it turns out that they are both derived from `LinearOp`, so all checks out.

How does this impact results?

benchmarks/scripts/expval_benchmark.py
natestemen: Using `H_p` doesn't seem ideal, but a good compromise. We should record the observable string _somewhere_ for easy reference.

Good find with that bug!

benchmarks/scripts/expval_benchmark.py
Misty-W: agree this is more readable, changed accordingly.

benchmarks/scripts/expval_benchmark.py
Misty-W: Good point. I added a comment:
` # Weights of edges between vertices and of the resulting unitary evolution`

natestemen: I mean more specifically where did the numbers come from?

Misty-W: The numbers are from the 10-vertex Barabasi-Albert graph in Fig 4(c) of the paper. I realized I need to add a docstring to `generate_qaoa_observable` anyway, so that's where I'll add the reference.

PR #266, opened 2025-02-26
Switch PyTket to use FullPeepHoleOptimize

jordandsullivan:
### Using PyTKET's FullPeepHoleOptimize
I was curious to go back and benchmark PyTKET's more robust FullPeepHoleOptimize compiler, which we did not use originally due to its being significantly slower (at the time ~50x ) than the other compilers. It is still much slower, but within an order of magnitude of Cirq, and it may be a fairer comparison for gate reduction performance.

The benchmark plots demonstrate a similar issue @bachase was referring to in #254 , where multiple runs of the same compiler on the same day will get messy. In this case, the three datapoints run today are with PyTKET with manual passes, and PyTKET using `FullPeepHoleOptimize`, first on my local laptop (different OS) then in the official GitHub runner.

We should be sure to differentiate these when loading the data in the next round of plot updates.  

**Compilation Time (relative to Cirq, Qiskit, and UCC)**

    Pytket =
        1.56× Cirq
        134.0× Qiskit
        55.2× UCC

**Compiled Multi-Qubit Gates Ratio (relative to Cirq, Qiskit, and UCC)**

    Pytket =
        0.968× Cirq
        0.927× Qiskit
        0.927× UCC

### Bottom line
I think we should switch to using FullPeepHoleOptimize for PyTKET. It looks like they are slightly outperforming us in terms of gate reduction (although it comes at the cost of 55x slower compile time). However we may be able to incorporate some PyTKET passes in a faster way using our https://ucc.readthedocs.io/en/latest/user_guide.html#writing-a-custom-pass guidelines.


jordandsullivan:
Latest commit cleans up plot to show only the Peephole data for PyTKET today:
image

jordandsullivan:
1. Yep, good idea. Created #269 , which includes a draft of the documentation section. 
2. We heard back from another maintainer and he felt the more lightweight manual version. wasn't a fair comparison to make, so I think we can just drop it for now.
 However, there is another "general purpose" PyTKET pass in addition to `FullPeepHoleOptimize` called **`KAKDecomposition`**, which gives the following performance:

image
As you can see, there is a new major version of PyTKET out, so I reran the benchmarks for this **`FullPeepHoleOptimize`** branch as well to check the performance:
image

Just to be clear, it definitely doesn't make sense to represent all these different implementations of PyTKET with the same color and label (manual version vs. FullPeephole vs. KAKDecomp). Going forward, as we display benchmarks, I would just drop the PyTKET data prior to when we start using FullPeephole or KAKDecomposition, and possibly include both of those in our benchmarks separately, since their performance is comparable.

3. We could have a less frequent benchmark for compilers which are much slower, since I imagine people might ask: "why didn't you compare against X compiler?" and we should probably have an answer for the biggest compilers out there. I might even create an issue template for it, like "Add a new compiler benchmark" where people can suggest new ones to add.

jordandsullivan:
For discussion on how to report the new data, see #272

benchmarks/scripts/common.py
```diff
@@ -80,24 +78,22 @@ def get_native_rep(qasm_string, compiler_alias):
...
-            DecomposeBoxes(),
-            RemoveRedundancies(),
-            AutoRebase({OpType.Rx, OpType.Ry, OpType.Rz, OpType.CX, OpType.H}),
-        ]
-    )
-    seqpass.apply(compilation_unit)
+    passes = [
+        FullPeepholeOptimise(),
+        AutoRebase({OpType.Rx, OpType.Ry, OpType.Rz, OpType.CX, OpType.H}),
+    ]
```
natestemen, line 95: Does it matter what order these are done in? I'm not sure if `FullPeepholeOptimise` retains the input gateset, but if so, then `AutoRebase` might make sense to act first.

benchmarks/scripts/common.py
```diff
@@ -61,8 +59,8 @@ def get_compile_function(compiler_alias):
     match compiler_alias:
         case "ucc":
             return ucc_compile
-        case "pytket":
-            return pytket_compile
+        case "pytket-peep":
+            return pytket_peep_compile
```
natestemen, line 63: For simplicity, does it make sense to keep this as simply "pytket"? If we are worried about clarity, then we could have part of our documentation that outlines exactly what the pytket/qiskit/cirq compilers are doing.

natestemen (commented):
`common.py` looks good! Just two questions.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -23,19 +23,29 @@ def extract_compiler_versions(header):
...
         header_line = f.readline().strip()
         compiler_versions = extract_compiler_versions(header_line)
+
     date_label = str(file).split("_")[1].split(".")[0]
     df = pd.read_csv(file, header=1)
+
     df["date"] = date_label
     df["compiled_ratio"] = df["compiled_multiq_gates"] / df["raw_multiq_gates"]
+    if "pytket-peep" in df["compiler"].unique():
+        compiler_versions["pytket-peep"] = compiler_versions["pytket"]
```
bachase, line 37: Just confirming I understand the logic here. Would this mean the file header had both `pytket` and `pytket-peep` versions in it?

bachase (commented):
Generally looks good, but also interested in @natestemen 's question regarding the order of the passes.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
jordandsullivan: No, it only records the pytket version in the header. This is just grabbing the Pytket version for plotting.

benchmarks/scripts/common.py
jordandsullivan: `FullPeepHole` has an optional argument which defaults to decomposing into CX gates:

Screenshot 2025-03-04 at 12 36 32 PM

The other options are TK1 and TK2, which I cannot find any references to outside of PyTKET itself: https://docs.quantinuum.com/tket/api-docs/circuit_class.html#pytket.circuit.Circuit.TK1

My vote is to keep the explicit rebase into our target gateset included after the call to be sure we're comparing apples to apples.

bachase: Perhaps as just a followup to this change-- is there any change in the resulting circuit properties if the rebase comes first versus second?

bachase (approved):
👍 given responses

benchmarks/scripts/common.py
jordandsullivan: Yeah. If you put it second, the compiled circuit contains those TK1 gates.

PR #267, opened 2025-02-26
Run pytest before ruff linter checks

jordandsullivan:
I want to test the functionality of a change before checking the formatting. In this way you get a failure like on an unused import before the code itself runs: https://github.com/unitaryfund/ucc/actions/runs/13554895918/job/37886926715

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,12 +31,12 @@ jobs:
 
       - name: Install dependencies & ucc
         run: poetry install
-
+        
+      - name: Run tests
+        run: poetry run pytest ucc --verbose
+        
       - name: Run linter
         run: poetry run ruff check
 
       - name: Run formatter check
         run: poetry run ruff format --check
-
-      - name: Run tests
-        run: poetry run pytest ucc --verbose


bachase (approved):
LGTM.

In addition, if you hadn't already, you could consider:
1. Depending on your IDE, you may be able to set the ruff format and ruff linter to run automatically whenever you save a file.
2. You can enable pre-commit hooks (mentioned in our docs [here](https://github.com/unitaryfund/ucc/blob/cd3d5b95ce4b0c25dab979e29933affb585098fe/docs/source/contributing.rst?plain=1#L40)) that will automatically format and fix any linter issues right before any calls to `git commit` complete (or will prevent commit if it would cause a conflict). The benefit there is it catches it earlier, and avoids an extra commit for fixing.

cc @natestemen and @Misty-W if they have any different workflows. We can consider documenting this more in the docs or README as well.

natestemen:
Okay by me. Another way to achieve this is to use [`continue-on-error`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error), so both the tests, and the linting can run. It does look like there is some uncertainty on the behavior of this flag from [stackoverflow](https://stackoverflow.com/a/73357322/21160270), but we could test it!

jordandsullivan:
I would rather just stick with the order for now rather than test a new flag. I will update my pre-commit hooks though thanks!

PR #290, opened 2025-03-05
Use target_device if specified, not just connectivity, when compiling

### LINKED ISSUE #288: Error when using custom `target_device`
**Describe the bug**
When attempting to specify an edge-list as the `target_device` argument to `compile`, the code errors with 

```
TranspilerError: 'A target must be specified or a coupling map and properties must be provided'
```

**To Reproduce**
Run the following code
```python
import cirq
from ucc import compile

qubits = cirq.LineQubit.range(3)
cirq_circuit = cirq.Circuit(
    cirq.H(qubits[0]),
    cirq.CNOT(qubits[0], qubits[1]),
    cirq.CNOT(qubits[0], qubits[2]) # <- 0 <> 2 not natively supported by topology
)
cirq_compiled_circuit = compile(cirq_circuit, target_device=[[0,1], [1,2]])
```

**Expected behavior**
I would expect the above code to run without error.

**Additional context**
This appears to be a consequence of #256, which removed a custom version of the `qiskit` `VF2PostLayout` class. In the custom version, the pass was changed to allow only specifying a coupling map without properties, as the optimization was using the number of multi-qubit gates in the optimization cost function. However, the `qiskit` versions uses gate error rates as part of cost function, so if a coupling map is provided, it also needs error properties specified. 

We could consider restoring the custom version, but that would lose any benefits of continued improvements in the `qiskit` version. `qiskit` also looks to be deprecating the coupling map + properties in favor of [`Target`](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.Target). 

As it stands, the interface to `ucc.compile` mixes these, taking in a `target_device` but only extracting its coupling map. Although this was intentional given the custom passes, that does break the user experience I would expect if I were to supply a full `target_device`.

The simplest fix for now is to only allow a `target_device` and pass that through to `UCCDefault1`.

bachase:
Fixes #288 by using the specified `target_device`, not just the coupling map, when provided to `compile` or `UCCDefault1`.

### DIFF
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -1,5 +1,3 @@
-from qiskit.providers import BackendV2
-from qiskit.transpiler import CouplingMap
 from qbraid.programs.alias_manager import get_program_type_alias
 from qbraid.transpiler import ConversionGraph
 from qbraid.transpiler import transpile
@@ -41,6 +39,7 @@ def compile(
         return_format (str): The format in which your circuit will be returned.
             e.g., "TKET", "OpenQASM2". Check ``ucc.supported_circuit_formats()``.
             Defaults to the format of the input circuit.
+        target_device (qiskit.transpiler.Target): (optional) The target device to compile the circuit for. None if no device to target
 
     Returns:
         object: The compiled circuit in the specified format.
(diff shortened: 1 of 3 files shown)


jordandsullivan (approved):
LGTM. I think actually keeping the abstraction on a higher level (passing the device itself rather than the coupling list) is better from an architecture design perspective, since we will ultimately want the ability to pass devices with more properties down to lower levels of the compiler stack.

PR #300, opened 2025-03-11
264 relabel qiskit

jordandsullivan:
Since #254 plans to remove the names of each compiler from the labels and only keep the version number (as the names are present in the legend), this PR is ready to go. Very simple change, but will look wonky until that label fix is in.


natestemen (approved):
Do we want to save the data under "qiskit-default" from here on out? Or is it simpler to keep the name the same and only change the label.

jordandsullivan:
I say keep saving it under just Qiskit for now and just change the label.

PR #301, opened 2025-03-12
Support custom passes in compile

### LINKED ISSUE #292: Support custom passes in `compile`
**Is your feature request related to a problem? Please describe.**

Currently, when using a custom set of the compilation passes, users lose the automatic translation that the default `compile` method provides

```python
from ucc import compile, UCCDefault1

compile(circuit) # circuit can be any supported frontend

ucc_compiler = UCCDefault1()
ucc_compiler.pass_manager.append(MyCustomPass()
ucc_compiler.run(circuit) # <--- this has to be the UCC "native" format, which is qiskit today
```

**Describe the solution you'd like**

As discussed in https://github.com/unitaryfund/ucc/discussions/291, the original plan was to instead use `compile` as the primary interface, where users can provide an optional set of passes if desired.

```python
compile(circuit) # compiler = None, so defaults to UCCDefault1

my_compiler = UCCDefault1()
my_compiler.pass_manager.append(MyCustomPass()

compile(circuit, compiler = my_compiler)
```

bachase:
Following the discussion in #292, adds a `custom_passes` argument to `compile`.

I also removed some now stale parts of the docs, but we should consider a separate issue/PR to add more details on the choice of passes in `UCCDefault`.
Resolves #292

### DIFF
--- a/docs/source/user_guide.rst
+++ b/docs/source/user_guide.rst
@@ -40,11 +40,16 @@ For example, we can define a random circuit in Qiskit and optimize it using the 
    gates = ["cx", "cz", "cy", "swap", "x", "y", "z", "s", "sdg", "h"]
    num_qubits = 10
    raw_circuit = random_clifford_circuit(
-       num_qubits, gates=gates, num_gates=10 * num_qubits * num_qubits
+      num_qubits, gates=gates, num_gates=10 * num_qubits * num_qubits
    )
    compiled_circuit = ucc.compile(raw_circuit)
-   print(f"Number of multi-qubit gates in original circuit: {count_multi_qubit_gates_qiskit(raw_circuit)}")
-   print(f"Number of multi-qubit gates in compiled circuit: {count_multi_qubit_gates_qiskit(compiled_circuit)}")
+   print(
+      f"Number of multi-qubit gates in original circuit: {count_multi_qubit_gates_qiskit(raw_circuit)}"
+   )
+   print(
+      f"Number of multi-qubit gates in compiled circuit: {count_multi_qubit_gates_qiskit(compiled_circuit)}"
+   )
+
 
 .. testoutput::
    :hide:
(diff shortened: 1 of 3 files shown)


ucc/compile.py
```diff
@@ -49,7 +48,10 @@ def compile(
 
     # Translate to Qiskit Circuit object
     qiskit_circuit = transpile(circuit, "qiskit")
-    compiled_circuit = UCCDefault1(target_device=target_device).run(
+    ucc_default1 = UCCDefault1(target_device=target_device)
+    if custom_passes is not None:
+        ucc_default1.pass_manager.append(custom_passes)
+    compiled_circuit = ucc_default1.run(
```
natestemen, line 54: Should the variable still be named `ucc_default1` if it gets modified and new passes get appended?

docs/source/user_guide.rst
```diff
@@ -40,11 +40,16 @@ For example, we can define a random circuit in Qiskit and optimize it using the
...
    )
    compiled_circuit = ucc.compile(raw_circuit)
-   print(f"Number of multi-qubit gates in original circuit: {count_multi_qubit_gates_qiskit(raw_circuit)}")
-   print(f"Number of multi-qubit gates in compiled circuit: {count_multi_qubit_gates_qiskit(compiled_circuit)}")
+   print(
+      f"Number of multi-qubit gates in original circuit: {count_multi_qubit_gates_qiskit(raw_circuit)}"
+   )
+   print(
+      f"Number of multi-qubit gates in compiled circuit: {count_multi_qubit_gates_qiskit(compiled_circuit)}"
+   )
```
natestemen, line 51: nit: Indentation with 3 spaces 😱.

docs/source/user_guide.rst
```diff
@@ -132,36 +125,39 @@ In the following example, we demonstrate how to create a custom pass, where the
 Applying a non-default pass in the transpilation sequence
 =========================================================
 
-UCC's built-in pass manager ``UCCDefault1().pass_manager`` can be used to apply a non-default or a custom pass in the sequence of transpilation passes.
-In the following example we show how to add passes for merging single qubit rotations interrupted by a commuting 2 qubit gate.
+The `compile` method accepts an optional list of custom passes to run after the default suite defined in the  built-in pass manager ``UCCDefault1().pass_manager``.
```
natestemen, line 128:
```suggestion
The ``compile`` method accepts an optional list of custom passes to run after the default suite defined in the  built-in pass manager ``UCCDefault1().pass_manager``.
```

docs/source/user_guide.rst
```diff
@@ -132,36 +125,39 @@ In the following example, we demonstrate how to create a custom pass, where the
...
-   from ucc import UCCDefault1
-   ucc_compiler = UCCDefault1()
-
-   ucc_compiler.pass_manager.append(MyCustomPass())
-
-   custom_compiled_circuit = ucc_compiler.run(circuit_to_compile)
+   from ucc import compile
+   custom_compiled_circuit = compile(
+      circuit_to_compile, custom_passes=[MyCustomPass()]
+   )
```
natestemen, line 160: Much cleaner workflow!

natestemen (approved):
Approving with some minor comments/suggestions.

docs/source/user_guide.rst
bachase: Yikes --- so embarrassing 😜 !!

bachase: Actually, double-embarrass perhaps. Looking again at existing `rst`, it looks like the code within codeblocks is currently indented with 3 spaces.

https://github.com/unitaryfund/ucc/blob/d92c3419a7d10dcdcb1bf29ef37662b7dc9dd3db/docs/source/user_guide.rst?plain=1#L12-L15

natestemen: Is that a standard `rst` thing?

bachase: The example in the [docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) has that pattern, but just testing, looks like as long as the indentation is consistent, the number of spaces doesn't matter.

If that's the case, I'll leave as is for now (where its 3 spaces throughout that file)

ucc/compile.py
bachase: No strong preference; it is still an instance of that class even if we are customizing it. Open to another (shorter =) ) name if you have one.

natestemen: Nothing comes to mind, and it's not blocking from me!

PR #309, opened 2025-03-14
Create first draft of python-publish.yml

jordandsullivan:
Add an automated workflow file for publishing to PyPI, which should run automatically upon a release. 
It includes a test upload to TestPyPI and then requires manual approval from @jordandsullivan , @Misty-W , or @bachase to publish to real PyPI. 

Need to merge into main to be able to test the manual dispatch function.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,58 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  release-build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Poetry
+        uses: abatilo/actions-poetry@v2
+        with:
+          poetry-version: "1.8.2" # Adjust to your required Poetry version
+
+      - name: Configure Poetry to use virtualenv inside project
+        run: poetry config virtualenvs.in-project true
+
+      - name: Install dependencies
+        run: poetry install --no-interaction --no-root
+
+      - name: Build release distributions
+        run: poetry build
+
+      - name: Upload distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+  pypi-publish:
+    runs-on: ubuntu-latest
+    needs:
+      - release-build
+    permissions:
+      id-token: write  # Enables Trusted Publishing (no need for API token)
+
+    environment:
+      name: pypi
+      url: https://pypi.org/project/ucc/${{ github.event.release.name }}
+
+    steps:
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+      - name: Publish release distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: dist/
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,4 +1,4 @@
-name: Upload Python Package
+name: Publish Python Package
 
 on:
   release:
@@ -6,6 +6,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write  # Required for OpenID Connect authentication with PyPI
 
 jobs:
   release-build:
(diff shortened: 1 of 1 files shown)


jordandsullivan:
Added this workflow file as a trusted publisher on PyPI. We just need to make sure this action can only be triggered by us maintainers.

Screenshot 2025-03-14 at 10 07 55 AM

Misty-W:
> We just need to make sure this action can only be triggered by us maintainers.

@natestemen, do you know how to set that up? I seem to remember you updated the Mitiq PyPI publishing workflow.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Build release distributions
+        run: poetry build
+
+      - name: Upload distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+  testpypi-publish:
```
natestemen, line 30: If this is going to run on every release, is there a reason to publish to testpypi right before pypi? Usually the testpypi publish step is run on a more regular basis to ensure nothing has broken. Otherwise, is it still needed?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  pypi-publish:
+    runs-on: ubuntu-latest
+    needs: testpypi-publish
+    if: github.ref == 'refs/heads/main'  # Ensures only main branch triggers PyPI deploy
+    permissions:
+      id-token: write
+    environment:
+      name: pypi
+      url: https://pypi.org/project/ucc/${{ github.event.release.name }}
```
natestemen, line 60: I can't find any docs on what type of URLs are accepted here. From https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/, I only see examples that end after `/project/ucc/`. Is what comes after a different way to specify the version?

natestemen (commented):
> We just need to make sure this action can only be triggered by us maintainers.

This should AFAIK already be taken care of by GitHub's restrictions as to who can create a release (maintainers). As per [GitHub docs](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository):

> Repository collaborators and people with write access to a repository can create, edit, and delete a release.

So as long as the maintainer list is kept up to date [here](https://github.com/unitaryfund/ucc/settings/access), then this should be safe.

jordandsullivan:
> This should AFAIK already be taken care of by GitHub's restrictions as to who can create a release (maintainers). 
> So as long as the maintainer list is kept up to date [here](https://github.com/unitaryfund/ucc/settings/access), then this should be safe.

Correct, that is my understanding.

.github/workflows/python-publish.yml
jordandsullivan: Yeah it just gets the release name which is typically the version: https://stackoverflow.com/questions/65927378/how-to-get-the-description-or-name-from-a-release-in-github-actions

.github/workflows/python-publish.yml
jordandsullivan: So this way it's set up, it will automatically push to TesPyPI and then manually require someone to approve it before it goes onto PyPI.

jordandsullivan: The idea is just for someone to make sure the test upload looks good and then manually push it to offical PyPI.

.github/workflows/python-publish.yml
natestemen: Ah yeah the `github.event.release.name` part makes sense, but I'm unsure about `pypi` accepting info after `ucc/`. I've never seen that.

Side note: does the `url` param need to be string-ified since there is a variable thrown in there?

.github/workflows/python-publish.yml
natestemen: Ah okay cool! That's a great idea. I can't quite figure out how that's working here. Can you help me understand?

.github/workflows/python-publish.yml
jordandsullivan: Ah actually the correct attribute is .tag_name rather than .name. Fixed that in latest commit. 

Yes, you can specify the version this way:
https://pypi.org/project/ucc/0.4.3/
https://pypi.org/project/mitiq/0.43.0/

I believe all you need is the ${{ }} delimiters.

.github/workflows/python-publish.yml
natestemen: Very nice setup! Thanks for the detailed explanation. All makes sense :)

jordandsullivan: Thanks, do you want to approve the PR?

natestemen (approved):

PR #311, opened 2025-03-19
Add manual trigger to PyPI publishing

jordandsullivan:
This will allow us to manually trigger a deployment to PyPI, in addition to automatically triggering upon a release. The relevant code change is [here](https://github.com/unitaryfund/ucc/pull/311/files#diff-87c8be2ac3b248aef84cc474af838d7cc461b7f8fb7f5444ac75f4d8fc02e377R63).

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -9,17 +9,6 @@ permissions:
   contents: read
 
 jobs:
-  debug:
-      runs-on: ubuntu-latest
-      if: (github.event.release.target_commitish == 'main' && github.event_name == 'release') || (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/jordandsullivan-patch-1')      
-      steps:
-        - name: Debug event data
-          run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -56,11 +56,11 @@ jobs:
 
       - name: Clean up distributions
         run: rm -rf dist/
-
+
```
natestemen, line 59: Oops!

natestemen (approved):
I often find one of the best ways to figure out oddities with GitHub actions is to look at other workflows on GitHub where someone is doing what I want. Is this suggestion based off something you found in the wild?

It's very hard to know if this achieves the goal without merging (even [`act`](https://github.com/nektos/act) I don't think will help here, right?), so I'll approve so as to not block, but with the caveat that I don't know if this will solve the issue.

jordandsullivan:
@natestemen yeah it's hard to know for sure before merging. I didn't find this combo specifically in the wild, since it includes environment configurations and a test step.

Issue #299, opened 2025-03-11
Make noise model more realistic


jordandsullivan:
Adjusting single and two qubit error rates to be more similar to the ratio in real hardware.

natestemen:
| qpu | 1q error rate | 2q error rate | ref |
| --- | --- | --- | --- |
| willow | $3.6\times 10^{-4}$ | $1.4\times10^{-3}$ | [spec sheet](https://quantumai.google/static/site-assets/downloads/willow-spec-sheet.pdf) |
| ibm eagle chips | NA | $2.11\times 10^{-2}$ | [IBM quantum platform](https://quantum.ibm.com/services/resources) |
| ibm heron chips | NA | $5.36\times 10^{-3}$ | ^ |
| zuchongzhi-3 | $1\times 10^{-3}$ | $3.8\times 10^{-3}$ | https://en.ustc.edu.cn/info/1007/5015.htm |
| ionq forte | $2\times 10^{-4}$ | $4\times10^{-3}$ | https://ionq.com/quantum-systems/compare |

**Average 1q error rate**: $5.2\times 10^{-4}$
**Average 2q error rate**: $7.1\times 10^{-3}$

PR #315, opened 2025-03-24
use error rates reflecting current quantum devices

natestemen:
## Description

Updates the single and two qubit error rates to be more representative of current hardware.

fixes https://github.com/unitaryfund/ucc/issues/299

### Reviewer notes

1. The representative error rates used in this PR are averages of what I was able to find online. I used average error rates instead of best where it was stated. The sample is small, but it does bring us much closer to current devices than what was previously running. My question is: does averaging these numbers make sense? Technically they become some "fake" device that doesn't exist (i.e. a combination of $5.2\times 10^{-4}$ 1q and $7.1\times 10^{-3}$ 2q error rates is maybe impossible). IMO this is negligible, but wanted to raise to see if others agree.
2. Comparing performance of compilers pre- and post-this change does not make sense as error rates will be drastically different. Should we delete the old data in this PR?


jordandsullivan:
Yeah, agreed. We can scrap it until we've gotten all the custom observables in.

benchmarks/scripts/expval_benchmark.py
```diff
@@ -21,8 +21,9 @@
 )
 from ucc import compile as ucc_compile
 
-SINGLE_QUBIT_ERROR_RATE = 0.01
-TWO_QUBIT_ERROR_RATE = 0.03
+# error rates representative of current hardware as of Mar 24, 2025
+SINGLE_QUBIT_ERROR_RATE = 0.00052
```
bachase, line 25: Is it possible to add hyperlinks or a reference to the sources used to determine these rates?

natestemen: WDYT of a link to the issue with the data? https://github.com/unitaryfund/ucc/issues/299#issuecomment-2748802174

bachase: 👍

natestemen:
@jordandsullivan / @bachase any further comments?

bachase (approved):
👍 good update to defaults.

jordandsullivan:
LGTM

Issue #227, opened 2025-02-13
Use argparse (or similar) for command line parsing


bachase:
**Describe the solution you'd like**
The benchmarking scripts currently parse arguments directly from sys. We should use [`argparse`](https://docs.python.org/3/library/argparse.html) to more robustly parse options, generate help etec.

https://github.com/unitaryfund/ucc/blob/c66a3e895d6f67b8e846ad45560ce909bf5ecf2d/benchmarks/scripts/benchmark_script.py#L11-L20

Shivansh20128:
Hi! I have made a PR for this issue. 
Thanks

bachase:
Thanks @Shivansh20128 . I left a comment in the PR but otherwise looking good.

In case you are looking at other issues, just a heads up that I've started prototyping #313, which will impact how many of the benchmarking issues linked in #235 will be addressed. So those might not be best to work on just now.

PR #319, opened 2025-03-27
Using argparse to parse arguments

Shivansh20128:
This PR adds the use of the argparse to parse command line arguments in the benchmarking scripts.

#Reslves #227

### DIFF
--- a/benchmarks/scripts/benchmark_script.py
+++ b/benchmarks/scripts/benchmark_script.py
@@ -1,5 +1,7 @@
 # benchmark_script.py
 import sys
+import argparse
+
 from common import (
     log_performance,
     save_results,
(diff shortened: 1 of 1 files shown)


benchmarks/scripts/benchmark_script.py
```diff
@@ -1,24 +1,37 @@
...
+args = parser.parse_args()
+
+# Get the QASM file, compiler, and results folder passed as command-line arguments
+qasm_file = args.qasm_file
+compiler_alias = args.compiler
+results_folder = args.results_folder
+
 # Ensure both QASM file and compiler are provided as arguments
-if len(sys.argv) < 4:
+if len(args) < 4:
```
bachase, line 29: Since the code is now using argparse, I think this check and printing usage is no longer needed. Would it be better to remove them?

Shivansh20128: Yes, I thought that too, but then I left it alone as there was no harm in keeping it. I will remove it

bachase (approved):
Looks good. Thanks for contribution 2!

PR #320, opened 2025-03-27
Update GitHub organization name from unitaryfund to unitaryfoundation

bachase:
In preparation of updating the Github organization name.


jordandsullivan (approved):
LGTM thanks for updating!

cosenal:
I put the metriq-gym equivalent PR in Draft mode, so there is no risk of merging it by mistake before the renaming is executed.

PR #321, opened 2025-03-31
Use normalized relative error for expectation value benchmarking

natestemen:
This PR updates the expectation value benchmarking plots to use a normalized relative error metric instead of absolute error. Relative error allows for a meaningful averaging across many circuit types and adding a stabilizing term (I use $\varepsilon = 0.1$) ensures we do not see huge relative errors when the ideal expectation value is close to 0.

fixes https://github.com/unitaryfoundation/ucc/issues/298


benchmarks/scripts/plot_expval_benchmarks_over_time.py
```diff
@@ -50,10 +50,23 @@
 # Filter on first occurrence of each compiler version based on the date
 df_all = df_all[df_all["date"].isin(unique_dates)]
 
-# Group by date and compiler, and calculate the average absolute error
+
+def normalize_relative_error(row):
+    """Calculate the relative error and normalize it to avoid division by zero."""
+    relative_error = abs(row["ideal_expval"] - row["expval"]) / max(
+        abs(row["ideal_expval"]), 0.1
```
bachase, line 57: More curious, what is the motivation to do the max of this versus just adding a smaller epsilon with no abs? e.g

```
epsilon = 1e-8
(df['ideal_expval'] - df['expval']).abs() / (df['ideal_expval'].abs() + epsilon)
```

bachase (approved):
Curious on one choice, but otherwise :+1:?

benchmarks/scripts/plot_expval_benchmarks_over_time.py
natestemen: > motivation to do the max of this versus just adding a smaller epsilon

Well, I hadn't considered that 😛. My mindset was to leave the relative error computation unchanged where possible, and using `max` achieved that. After checking the relative errors in the [most recent benchmark run](https://github.com/unitaryfoundation/ucc/blob/ddf8d48b7bc52398119dabe8bd5a8139dd0b5a7e/benchmarks/results/expval_2025-03-04_21.csv) which range between `1e-14` and `1e-18`, I think your suggestion is very reasonable. Added in https://github.com/unitaryfoundation/ucc/pull/321/commits/9c9a655b42096e4966f592f7fce206cb1aae7535.

bachase (approved):

PR #322, opened 2025-03-31
re-enable expval benchmarks

natestemen:
Should fix the issue arising on [CI](https://github.com/unitaryfoundation/ucc/actions/runs/14178573065/job/39719233207) since we deleted the expectation value benchmarking data in https://github.com/unitaryfoundation/ucc/pull/315


jordandsullivan:
Can you add a plot here so we can take a look?

natestemen:
Good idea! I ran all the plotting functionality we have for expval benchmarking and pushed the results+plots in https://github.com/unitaryfoundation/ucc/pull/322/commits/ce26d86b3846b2796fae3c427b79bd5310b2b91d. I can remove before merging if that's appropriate.

jordandsullivan:
I thought we we had agreed on plotting _relative_ errors for the average rather than absolute errors? Unless it would be a lot of extra work, can you add a plot for that @natestemen ?

jordandsullivan:
Thanks -- it looks like we're still missing pytket-peep on the plot though:
image

natestemen:
Good catch. I ported over the fix from https://github.com/unitaryfoundation/ucc/pull/266 when the peephole optimization was added.

jordandsullivan (approved):
LGTM thanks!

PR #324, opened 2025-04-03
fix expval benchmark plotting script name

natestemen:
Missed this change in https://github.com/unitaryfoundation/ucc/pull/321. oops!


bachase (approved):
👍

PR #325, opened 2025-04-03
Update dependencies (including to qiskit 2.0.0)

bachase:
Updates current dependencies, in particular to qiskit 2.0.0 which required one change to the code.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -313,75 +313,75 @@ test-no-images = ["pytest", "pytest-cov", "pytest-rerunfailures", "pytest-xdist"
 
 [[package]]
 name = "coverage"
-version = "7.7.1"
+version = "7.8.0"
 description = "Code coverage measurement for Python"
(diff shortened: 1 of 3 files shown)


bachase:
Looks like a degradation in performance with qiskit.2.0.0 for the benchmark circuits.
latest by time

bachase:
@Misty-W -- I'm not too sure. The [release notes](https://docs.quantum.ibm.com/api/qiskit/release-notes/2.0) for qiskit 2.0.0 release is lengthy! Perhaps worth us taking a closer look at somepoint?

bachase:
@Misty-W I want to make one more change and will request a re-review before then (so don't look quite yet). But realizing this makes ucc incompatibile with users that might have qiskit < 2.0.0. I want to fix the code to support both, and then consider adding a maximum compatible version in our pyproject.toml.

PR #327, opened 2025-04-04
Add workflow to trigger ucc-bench run

### LINKED ISSUE #326: Add github actions to trigger ucc-bench benchmarks
As part of #313, I am developing a separate repository (unitaryfoundation/ucc-bench) to manage configuration and storing benchmark results. But pushes in the `ucc` repo will still need to trigger changes in that downstream repo for the benchmarks to run.

For this task, we are focusing on two triggers.

1. Trigger a benchmark run in `ucc-bench` in response to a commit to main `ucc`. This new workflow will trigger the [upgrade_ucc action[(https://github.com/unitaryfoundation/ucc-bench/blob/main/.github/workflows/upgrade_ucc.yml) in ucc-bench, whenever there is a push to main in ucc.
2. Trigger a benchmark preview run in `ucc-bench` for a PR opened in `ucc` with the `preview-benchmark-results` label. This new workflow will trigger the [run_benchmark_pr_ucc action[(https://github.com/unitaryfoundation/ucc-bench/blob/main/.github/workflows/run_benchmarks_pr_ucc.yml) in `ucc-bench`.

bachase:
This workflow will trigger a benchmark run in the new `ucc-bench` repo on a merge to main in `ucc`. 

This leverage [`repository_dispatch`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#repository_dispatch), where the action in the `ucc` repo calls a rest API to trigger an action in `ucc-bench`. The permissions were managed by creating a GitHub APP, installing that app into the `ucc-bench` repo with proper permissions, and then adding the app's key as secrets to the `ucc` repo.

We can run the existing benchmarking code within `ucc` and the new code in `ucc-bench` in parallel and then chose to cut-over when ready. The `ucc-bench` code is still finalizing, but I would like to enable this trigger sooner to test it, but then also start capturing the benchmark results as `ucc` development continues.

Resolve #326

### DIFF
--- a/.github/workflows/trigger-ucc-bench.yml
+++ b/.github/workflows/trigger-ucc-bench.yml
@@ -1,0 +1,36 @@
+# Trigger an upgrade of ucc version  in ucc-bench repo in response to push to main
+# To trigger an event in the other repo, we had to install a GitHub app in the
+# ucc-bench repo, and then add the app ID/private key as secrets to this repo
+# so it could trigger the workflow. See
+# https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions
+# for details.
+
+name: Trigger ucc upgrade in ucc-bench
+on:
+  push:
+    branches:
+      - main
+jobs:
+  trigger-benchmarks:
+    runs-on: ubuntu-latest
+    # Don't run on forks or original benchmark commits
+    if: ${{ github.repository == 'unitaryfoundation/ucc' && !contains(github.event.head_commit.message, '[benchmark chore]') }}
+
+    steps:
+      - name: Generate token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.UCC_BENCH_APP_ID }}
+          private-key: ${{ secrets.UCC_BENCH_APP_PRIVATE_KEY }}
+
+      - name: Trigger ucc-main-push event in ucc-bench
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+        run: |
+          gh api \
+            --method POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/unitaryfoundation/ucc-bench/dispatches \
+            -f "event_type=ucc-main-push" -F "client_payload[commit_hash]=${{ github.sha }}"


jordandsullivan:
Thanks Brad, this is looking really good. It looks like we're maybe missing compiled/uncompiled gates ratio in the table.

bachase:
> Thanks Brad, this is looking really good. It looks like we're maybe missing compiled/uncompiled gates ratio in the table.

Can add (was doing some final tweaks). Would you want that and the absolute gate counts? Or just go with the ratio instead?

bachase:
Sorry for spam here -- need to debug the comments a bit more.

bachase:
Ok, think it's finally ready for review. But likely easier after I give an overview of ucc-bench.

.github/workflows/trigger-ucc-bench-pr.yml
```diff
@@ -0,0 +1,51 @@
...
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
+        run: |
+          gh api \
+            --method POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/unitaryfoundation/ucc-bench/dispatches \
+            -f "event_type=ucc-main-pr" -F "client_payload[commit_hash]=${{ github.sha }}"\
+            -F "client_payload[pr_number]=${{ github.event.pull_request.number }}"
```
natestemen, line 51: This `repository_dispatch` feature is pretty cool! Glad to know about it.

Is it possible that some actor outside of this github action could mock the POST request to trigger the run? I guess the `GH_TOKEN` maybe prevents that?

benchmarks/scripts/plot_avg_benchmarks_over_time.py
```diff
@@ -66,7 +68,8 @@
 
 # Filter on first occurrence of each compiler version based on the date
 df_dates = df_dates[df_dates["date"].isin(unique_dates)]
-
+if SAVE_LEGACY:
+    df_dates.to_csv("./legacy_timing_results.csv", index=False)
```
natestemen, line 72: When would we want to set `SAVE_LEGACY = True`?

natestemen (approved):
Looks great! The context from the demo makes this pretty easy to understand.

I have two non-blocking questions :)

.github/workflows/trigger-ucc-bench-pr.yml
bachase: Yes that is correct. They would need to generate a token that has the proper access permissions on the ucc-bench repo. 

On the benefits side, users with that access (say with write access on the ucc-bench repo) could test that workflow via this call themselves.

benchmarks/scripts/plot_avg_benchmarks_over_time.py
bachase: I did this just once and manually imported the results here https://github.com/unitaryfoundation/ucc-bench/blob/main/results/ucc-benchmarks-8-core-U22.04/compilation_benchmarks/legacy_timing_results.csv.

We shouldn't need to redo that, so I could probably just drop this change. But was convenient enough to keep and I expect us to remove these benchmark scripts once we flip over to `ucc-bench`.

PR #329, opened 2025-04-09
Reduce single-qubit gates introduced by `ConsolidateBlocks` pass in result circuit

### LINKED ISSUE #302: Unexpected behavior with default compilation set
**Describe the bug**
While reviewing https://github.com/unitaryfund/ucc-demo/pull/1, @jordandsullivan [noticed](https://github.com/unitaryfund/ucc-demo/pull/1#issuecomment-2716034212) that a simple Bell state circuit was transforming the circuit even though it was already in the target gate basis:

Image

Debugging a bit, this appears to be due to forcing consolidation of blocks, after which unitary synthesis expands to the different gateset:

https://github.com/unitaryfund/ucc/blob/d92c3419a7d10dcdcb1bf29ef37662b7dc9dd3db/ucc/transpilers/ucc_defaults.py#L73

**To Reproduce**
1. Run the circuit in the unit tests https://github.com/unitaryfund/ucc/blob/d92c3419a7d10dcdcb1bf29ef37662b7dc9dd3db/ucc/tests/test_compile.py#L19-L22

To debug, I edited `UCCDefault1.run` to print the circuit after each pass:
```python
def callback_func(**kwargs):
    from qiskit.converters import dag_to_circuit

    pass_ = kwargs["pass_"]
    dag = kwargs["dag"]
    print(f"Pass: {pass_.name()}")
    print(dag_to_circuit(dag))
    print("-" * 80)

class UCCDefault1:
....
   def run(self, circuits):
       return self.pass_manager.run(circuits, callback=callback_func)

```

**Expected behavior**
We would expect the circuit to stay the same.

Misty-W:
Fixes #302. 

The reason why `ConsolidateBlocks` with force_consolidate = True` gave a lengthy decomposition of single-qubit gates, e.g. in the case of the Bell circuit, is that `force_consolidate=True` forces decomposition as well as cancellation, and our subsequent default passes didn't recognize those gates could be cancelled.

We get an improvement in 2-qubit gate count on QFT if we keep the original implementation where `force_consolidate` forces decompositions, so I changed default back to using the Qiskit implementation for now. 

Adding (uncommenting) Optimize1qGatesSimpleCommutation and a basis translation step improves the Bell circuit but doesn't completely restore it, even though H is in our target basis.

image


Misty-W:
@jordandsullivan, @bachase, I ran the benchmarking GitHub workflow on this branch `302-unexpected-behavior-with-default-compilation-set`, and the compile time increased due to the addition of the `Optimize1qGatesSimpleCommutation` pass, while two-qubit gate counts remained the same.
image

Including the `Optimize1qGatesSimpleCommutation` pass improved the result on the Bell circuit, but didn't fully restore it, even though H is in our target basis.
image

Misty-W:
Decision (in UCC team mtg): close without merging per previous comment.

PR #331, opened 2025-04-11
Fix pytket version ranges

bachase:
#325 accidentally restricted pytket to `<2.0.0`, when we had already upgraded to the `2.0` line. This fixes to restrict to `<3.0.0`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -199,30 +199,30 @@ files = [
 
 [[package]]
 name = "cirq-core"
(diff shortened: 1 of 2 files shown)


bachase:
I also cancelled the run for https://github.com/unitaryfoundation/ucc/actions/runs/14403888454 which would otherwise downgrade `pytket` in the official runs.

natestemen (approved):

Issue #280, opened 2025-03-04
Add documentation on benchmark circuits


bachase:
Add documentation on the circuits in the benchmark suite,  how they are defined and what they are meant to assess. This is to provide more context and background information, and a structure for any future circuits we would add.

PR #335, opened 2025-04-18
Updated documentation and README to point to ucc-bench

### LINKED ISSUE #269: Add compiler function details to docs
**Per discussion in #266**

**We want to add a brief explanation in the benchmarking section of our docs (with links to the source code) for each compiler. Something along the lines of:**

Our goal is to benchmark quantum compilers with settings as close as possible, so we define compile functions to capture these settings.*
This means that each compile function must:
1. Decompose to the same target gateset (specifically we chose, RX, RY, RZ, H, CX)
_Why?_ 
Our benchmarks compare the number of multi-qubit gates before and after compilation, so we need to make sure we are comparing apples to apples. In addition, when running on a real quantum device, only certain gates will be natively supported.
2. Require little manual user optimization (i.e. it is not a custom set of transpilation passes for a specific use case, but a general purpose pass). 
_Why?_
A core aspect of UCC's design philosophy is to make compiling easy and painless for people running quantum algorithms, and not require the user to have in-depth knowledge of compilation to get solid performance.

Here are the specific settings we use for each compiler:
- **`ucc_compile`**:
```Python
ucc.compile(circuit)
```
which by default returns an optimized circuit decomposed into the gateset [RX, RY, RZ, H, CX], unless a `target_device` is specified. 
- **`qiskit_compile`**: 
```Python
qiskit.transpile(circuit, optimization_level=3, basis_gates=["rz", "rx", "ry", "h", "cx"])
```
which is the maximum optimization level and by far the most common setting people use in code examples.

- **`pytket_peep_compile`**: 
```Python
def pytket_peep_compile(pytket_circuit):
    compilation_unit = CompilationUnit(pytket_circuit)
    passes = [
        FullPeepholeOptimise(),
        AutoRebase({OpType.Rx, OpType.Ry, OpType.Rz, OpType.CX, OpType.H}),
    ]
    SequencePass(passes).apply(compilation_unit)
    return compilation_unit.circuit
``` 
There are two different "general purpose circuit optimizers" in the [PyTKET documentation](https://docs.quantinuum.com/tket/api-docs/passes.html#pytket.passes): `FullPeepholeOptimise` and `KAKDecomposition`. Because both of these have comparable performance and the same call structure, we include both in our benchmarks.

- **`pytket_kak_compile`**: 
```Python
def pytket_peep_compile(pytket_circuit):
    compilation_unit = CompilationUnit(pytket_circuit)
    passes = [
        KAKDecomposition(),
        AutoRebase({OpType.Rx, OpType.Ry, OpType.Rz, OpType.CX, OpType.H}),
    ]
    SequencePass(passes).apply(compilation_unit)
    return compilation_unit.circuit
``` 

- **`cirq_compile`**: 
```Python
optimize_for_target_gateset(
        cirq_circuit, gateset=BenchmarkTargetGateset()
    )
```
where we have manually defined `BenchmarkTargetGateset` again as [RX, RY, RZ, H, CX]. Cirq doesn't have a default compile function precisely (as far as the UCC team can tell), but optimizing for a specific gateset is possible.

*If you disagree with us about how a compiler is being compared, please create an issue for it! We want our benchmarks to be useful and fair.

bachase:
Resolves #269 and #280.

We can wait to merge until after we decide to flip the switch to `ucc-bench`.


docs/source/benchmarking.rst
```diff
@@ -26,34 +27,75 @@ Our benchmarks follow a structured methodology to ensure consistency and reliabi
...
-:doc:`contributing guide <contributing>` to setup your dev environment. Then, run
-the benchmark script via:
+`ucc-bench README <https://github.com/unitaryfoundation/ucc-bench/blob/main/README.md>`_ to clone the repository and setup your environment.
+
+You then run the benchmark script via:
 
 .. code-block:: sh
 
-   poetry run ./benchmarks/scripts/run_benchmarks.sh <num_parallel>
+   uv run ucc-bench
```
jordandsullivan, line 51: I wonder if we should remove this snippet and just direct people to check out the instructions on ucc-bench. If you weren't reading carefully and just copied the snippet, you might try to run this in ucc itself and get confused.

docs/source/benchmarking.rst
```diff
@@ -26,34 +27,75 @@ Our benchmarks follow a structured methodology to ensure consistency and reliabi
...
-``benchmarks/results/`` as CSV files.
+- **Qiskit** is configured with ``optimization_level`` 3.
+- **Cirq** relies on a custom written ``BenchmarkTargetGateset`` to target the above mentioned basis gates.
+- **PyTKET** uses the  "`Full Peephole <https://docs.quantinuum.com/tket/api-docs/passes.html#pytket.passes.FullPeepholeOptimise>`_" optimization strategy.
+- **PyQPanda3** uses optimization level 2. Note that PyQPanda3 is not open source.
 
-The script assumes you have GNU parallel available on your machine, available
-via ``apt-get install parallel`` on Ubuntu or ``brew install parallel`` on MacOS.
 
 Contributing to benchmarks
```
jordandsullivan, line 100: I wonder if for this section, it would be better to just direct people to make an issue in ucc-bench. Again if someone weren't reading closely (since the repo names are pretty close), they might just try to create an issue in ucc itself.

README.md
```diff
@@ -77,14 +77,15 @@ def test_cirq_compile():
...
-![alt text](benchmarks/latest_compiler_benchmarks_by_circuit.png)
+![alt text](https://github.com/unitaryfoundation/ucc-bench/blob/main/results/ucc-benchmarks-8-core-U22.04/latest_compiler_benchmarks_by_circuit.png?raw=true)
 
 And here you can see progress over time, with new package versions labeled for each compiler:
-![alt text](benchmarks/avg_compiler_benchmarks_over_time.png)
-where pytket-peep indicates the `FullPeepHoleOptimize` function of PyTKET. Note that the compile times before 2024-12-10 may have been run on different classical compute instances, so their runtime is not reported here, but you can find this data in benchmarks/results.
-After 2024-12-10, all data present in this plot is on the same compute instance using our [ucc-benchmarks](https://github.com/unitaryfoundation/ucc/blob/main/.github/workflows/ucc-benchmarks.yml) GitHub Actions workflow.
+![alt text](https://github.com/unitaryfoundation/ucc-bench/blob/main/results/ucc-benchmarks-8-core-U22.04/avg_compiler_benchmarks_over_time.png?raw=true)
```
jordandsullivan, line 83: Elegant solution!

jordandsullivan (approved):
Looks great. Thanks for doing this re-architecting work @bachase ! Super valuable to have this reproducibility and modularity for the benchmarking suite.

docs/source/benchmarking.rst
bachase: Good suggestion. I think thats a better balance. I'll update.

docs/source/benchmarking.rst
bachase: Good suggestion as well. I'll update to be clearer.

docs/source/benchmarking.rst
```diff
@@ -99,4 +99,4 @@ We welcome contributions to improve UCC’s benchmarking suite! This could inclu
 - **Improving Benchmark Metrics** to add additional metrics of interest to compare compiler performance.
 - **Optimizing Compiler Configurations** to improve the default configuration of compilers in the benchmark.
 
-To contribute, please open an issue in the `ucc-bench <https://github.com/bachase/ucc-bench>_` repository.
+To contribute, please open an issue in the `ucc-bench <https://github.com/bachase/ucc-bench>`_ repository.
```
bachase, line 102: Thanks @jordandsullivan  --- rst syntax always escapes me!

jordandsullivan: All good I fixed it with Co-pilot :) 
Looks like it somehow broke the tests though 🤔

bachase: Ah yes, hit that one earlier when I was originally writing

```
/home/runner/work/ucc/ucc/docs/source/benchmarking.rst:4: WARNING: Duplicate explicit target name: "ucc-bench". [docutils]
```

Tests run `make docs` in strict mode. This one is complaining because there are two different links with text `ucc-bench`, but different URLs. Sphinx warns because you might've wanted both targets (which in this case is the text of the link) to point to the same thing. Will fix

bachase: And the "fix' is to make it a double-underscore vs single underscore after the link to tell sphinx/rst to keep the target anonymous vs tracking it.

PR #347, opened 2025-04-28
Migrate to uv for package/project management

### LINKED ISSUE #345: Switch to uv for package management
As part of #313 and in developing `ucc-bench`, I found that `uv` is working well, including with dependabot and in github action workflows and for a variety of reasons mentioned in https://github.com/unitaryfoundation/mitiq/discussions/2674. We should make the same switch for `ucc` for consistency.

This should be relatively painless, requiring
1. Updating the pyproject.toml to standard PEP stanzas versus `tool.poetry` ones
2. Updating the CI/CD workflows to use `uv` over poetry (following examples in ucc-bench)
3. Updating the documentation
4. Updating dependabot to use uv

bachase:
Resolves #345 .  Given the good experience using uv in `ucc-bench`, doing the migration here as well.

For current developers, after this gets merged, you would need to ensure `uv` is installed, and then run

```
uv sync --all-extras --dev
```

to get the relevant packages installed. As mentioned in the docs, you would then use `uv run ...` in place of `poetry run ...`


docs/source/contributing.rst
```diff
@@ -15,17 +15,17 @@ The rest of this document describes the technical details of getting set up to d
...
 
     git clone https://github.com/unitaryfoundation/ucc.git
     cd ucc
-    poetry install
+    uv sync --all-extras --all-groups
 
-For all of the following commands, we assume you either prefix each command with ``poetry run``, or
-you first active the `poetry managed virtual environment <https://python-poetry.org/docs/managing-environments/#activating-the-environment>`_ by running the output of ``poetry env activate`` in your shell.
+For all of the following commands, we assume you either prefix each command with ``uv run``, or
+you first active the `uv managed virtual environment <https://docs.astral.sh/uv/pip/environments/#using-a-virtual-environment>`_ by running ``source .venv/bin/activate`` in your shell.
```
bdg221, line 28: you first **activate** the ...

This will match the [README](https://github.com/unitaryfoundation/ucc/blob/9c8430e0b6eb63d280d30539b76db1bfde44fda2/README.md?plain=1#L40)

.github/workflows/test.yml
```diff
@@ -16,31 +16,28 @@ jobs:
...
-      - name: Install Poetry
-        run: |
-          python -m pip install --upgrade pip
-          pip install poetry
+          python-version-file: "pyproject.toml"
 
-      - name: Install dependencies & ucc
-        run: poetry install
+      - name: Install the project
+        run: uv sync --all-extras --all-groups
```
jordandsullivan, line 30: I find this install command pretty unintuitive. Most people in quantum are used to `pip install .` 

Can we alias this uv command with something simpler?

.github/workflows/python-publish.yml
```diff
@@ -56,10 +62,10 @@ jobs:
 
       - name: Clean up distributions
         run: rm -rf dist/
-  
+
   pypi-publish:
     runs-on: ubuntu-latest
-    needs: testpypi-publish # If statement guarantees we are on main and either in a release or manual dispatch event. 
+    needs: testpypi-publish # If statement guarantees we are on main and either in a release or manual dispatch event.
```
jordandsullivan, line 68: Have you tested this new uv infrastructure with the TestPyPI workflow?

README.md
```diff
@@ -29,16 +29,13 @@ By leveraging [qBraid](https://github.com/qBraid/qBraid), UCC interfaces automat
...
-If developing, please install [Poetry](https://github.com/python-poetry/install.python-poetry.org), which is used to managed dependencies. Then setup a dev version via:
+If developing, please install [uv](https://docs.astral.sh/uv/getting-started/installation/), which is used to managed dependencies. Then setup a dev version via:
 
 ```bash
 git clone https://github.com/unitaryfoundation/ucc.git
 cd ucc
-poetry install
+uv sync --all-extras --all-groups
 ```
-For any subsequent commands mentioned in the docs, we assume you either prefix each command with `poetry run`, or
```
jordandsullivan, line 39: Let's copy the info about uv environments from the contributing guide to also be present here in the README

jordandsullivan (commented):
As a high level comment, what is the motivation for switching to uv? From a user perspective, most quantum folks will not be familiar with it, and I'm not sure I understand the advantage from a dev and maintenance perspective.

.github/workflows/test.yml
bachase: Users of `ucc` would still be recommended to install via `pip install ucc`. `uv` is used here to ensure we are testing that in a specific environment, where the specific set of packages and dependencies is reproducible for others who are developing (and as represented in `uv.lock`. The extra logic here is because even for developers, some packages are optional (e.g. sphinx and docs ones, or formatting ones like `ruff`).

Since this github workflow runs unit tests, runs linter checks and does a "smoke" test of the docs build, it needs all of those dependencies.

README.md
bachase: Can do. But also, per your comments above, is there another way to stress these steps are for developers vs library users?

.github/workflows/python-publish.yml
bachase: Not yet. Is the right way to manually run https://github.com/unitaryfoundation/ucc/actions/workflows/python-publish.yml on this branch, but stop after the test pypi step?

bachase: Trying that now https://github.com/unitaryfoundation/ucc/actions/runs/14717479687/job/41304148703

bachase: Got an error (https://github.com/unitaryfoundation/ucc/actions/runs/14717479687/job/41304165076), but this appears to be the same one @Misty-W had when doing the prior release (https://github.com/unitaryfoundation/ucc/actions/runs/14508847696/job/40703419908). Will investigate that

README.md
jordandsullivan: Many users will want to clone the repo and install it locally, esp as we intend for one of our main user groups to be developers of Transpiler passes.

.github/workflows/python-publish.yml
bachase: Running on a dedicated [branch](https://github.com/unitaryfoundation/ucc/compare/main...bc-test-pypi) without these `uv` changes but with `verbose` output, the [error](https://github.com/unitaryfoundation/ucc/actions/runs/14717595606/job/41304547378) looks to be because an existing upload for that release number already exists.

@jordandsullivan -- do you think having the test workflow operational for non-releases is important? If so, I suspect we would need to modify the upload filename/artifact to have some pre-release identifiers so they don't clash.

For this `uv` change, I believe all the build releated steps from `uv` worked properly, so should be clear and we can track anything else on publishing as a separate issue.

.github/workflows/test.yml
jordandsullivan: In general I would say our users are likely to be at least code developers, so it makes sense to consider the ease of their experience. IMO we should be cautious when introducing infrastructure complexity that may make it harder for these developers to contribute, since they may not be familiar with these tools. I haven't used uv previously myself, and I am trying to understand the advantage we gain from switching to it over poetry, from a maintainer and from a user perspective.

.github/workflows/python-publish.yml
jordandsullivan: I think it's not critical for every issue, but IMO we should be able to do a manual trigger test to PyPI for issues that deal with build-related changes, so I think it's worth fixing the tag collision separately. Doesn't need to be a blocker for this issue though since the errors appear to be coming from elsewhere.

.github/workflows/test.yml
bachase: Got it. 

We discussed some of the advantages over poetry in this Mitiq [discussion](https://github.com/unitaryfoundation/mitiq/discussions/2674) . Since that discussion, dependabot now support `uv` natively. In fact, the `poetry` support in dependabot is [broken](https://github.com/dependabot/dependabot-core/issues/11237) at the moment. I think those are all fairly compelling for developers, much better supported today. Not necessarily the best argument, but others in the ecosystem are also using it, e.g. [Kirin](https://github.com/QuEraComputing/kirin).

But I agree on complexity creep. In this case, I think this actually saves us there too, because `uv` does a better job then `poetry` for the reasons in its [homepage](https://docs.astral.sh/uv/). 

Perhaps to combat, we should link to some uv tutorials? Or add a section in the docs on using `uv` for development with some example workflows? We can also host a session on discord on the tool if there's broader interest.

.github/workflows/python-publish.yml
bachase: I'll open an issue on it.

bachase: https://github.com/unitaryfoundation/ucc/issues/350

.github/workflows/test.yml
jordandsullivan: Got it, thanks for the impact summary! I agree, those are compelling reasons to switch from a dev perspective. 

I think including a link to uv docs/tutorials would be nice the instructions for setting up an environment and install. 

I bet there might be some broader interest in sharing the uv tool with other UF folks in the context of our work :)

jordandsullivan (approved):
Pending docs updates as discussed, LGTM!

bachase:
Thanks for all the feedback. Added some additional details on `uv`, and added links to its documentation and good recent tutorial.

PR #354, opened 2025-05-01
342 target gateset

jordandsullivan:
The fix I went for was to allow users to pass 
1) a `return_gateset` in the form of `dict` keys (or a set) and use the `qiskit.transpile(circuit, basis_gates={'my', 'gate', 'set'}, optimization=0)` functionality to convert to a new basis automatically OR
2) a `target_device` whose allowable operations will be checked if that is an attribute present in that device. 

Added tests to confirm that the user-defined gateset is respected when it is passed, and the same for the target device.

**Note:** there is currently no functionality to check that a user-defined gateset is a) universal or b) available on the target_device, but currently these will cause errors that are reasonably straightforward to interpret for the user ('this operation is not available on this target'), so I think we can leave that for a future issue.

### DIFF
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -28,7 +28,7 @@ supported_circuit_formats = ConversionGraph().nodes()
 def compile(
     circuit,
     return_format="original",
-    return_gateset=None,
+    target_gateset=None,
     target_device=None,
     custom_passes=None,
 ):
@@ -41,7 +41,7 @@ def compile(
         return_format (str): The format in which your circuit will be returned.
             e.g., "TKET", "OpenQASM2". Check ``ucc.supported_circuit_formats()``.
             Defaults to the format of the input circuit.
-        return_gateset (set[str]): (optional) The gateset to compile the circuit to.
+        target_gateset (set[str]): (optional) The gateset to compile the circuit to.
             e.g. {"cx", "rx",...}. Defaults to the gateset of the target device, or if none is provided, {"cx", "rz", "rx", "ry", "h"}.
         target_device (qiskit.transpiler.Target): (optional) The target device to compile the circuit for. None if no device to target
         custom_passes (list[qiskit.transpiler.TransformationPass]): (optional) A list of custom passes to apply after the default set
@@ -61,20 +61,19 @@ def compile(
         qiskit_circuit,
     )
 
-    if return_gateset is not None:
+    if target_gateset is not None:
         # Translate into user-defined gateset; no optimization
-        # TODO: Check if the gateset is valid for the target device
         compiled_circuit = qiskit_transpile(
-            compiled_circuit, basis_gates=return_gateset, optimization_level=0
+            compiled_circuit, basis_gates=target_gateset, optimization_level=0
         )
     else:
         try:
             # Use target_device gateset if available
-            return_gateset = target_device.operation_names
+            target_gateset = target_device.operation_names
             # Translate into the target device gateset; no optimization
             compiled_circuit = qiskit_transpile(
                 compiled_circuit,
-                basis_gates=return_gateset,
+                basis_gates=target_gateset,
                 optimization_level=0,
             )
         except AttributeError:
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -66,18 +66,16 @@ def compile(
         compiled_circuit = qiskit_transpile(
             compiled_circuit, basis_gates=target_gateset, optimization_level=0
         )
-    else:
-        try:
-            # Use target_device gateset if available
-            target_gateset = target_device.operation_names
-            # Translate into the target device gateset; no optimization
-            compiled_circuit = qiskit_transpile(
-                compiled_circuit,
-                basis_gates=target_gateset,
-                optimization_level=0,
-            )
-        except AttributeError:
-            pass  # Use UCC default gateset
+    elif hasattr(target_device, "operation_names"):
+        # Use target_device gateset if available
+        target_gateset = target_device.operation_names
+
+        # Translate into the target device gateset; no optimization
+        compiled_circuit = qiskit_transpile(
+            compiled_circuit,
+            basis_gates=target_gateset,
+            optimization_level=0,
+        )
 
     # Translate the compiled circuit to the desired format
     final_result = translate(compiled_circuit, return_format)
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -67,6 +67,10 @@ def compile(
             compiled_circuit, basis_gates=target_gateset, optimization_level=0
         )
     elif hasattr(target_device, "operation_names"):
+        if target_gateset not in target_device.operation_names:
+            warnings.warn(
+                f"Warning: The target gateset {target_gateset} is not supported by the target device. "
+            )
         # Use target_device gateset if available
         target_gateset = target_device.operation_names
 
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -63,6 +63,7 @@ def compile(
 
     if return_gateset is not None:
         # Translate into user-defined gateset; no optimization
+        # TODO: Check if the gateset is valid for the target device
         compiled_circuit = qiskit_transpile(
             compiled_circuit, basis_gates=return_gateset, optimization_level=0
         )
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -41,7 +41,7 @@ def compile(
         return_format (str): The format in which your circuit will be returned.
             e.g., "TKET", "OpenQASM2". Check ``ucc.supported_circuit_formats()``.
             Defaults to the format of the input circuit.
-        return_gateset (set): (optional) The gateset to compile the circuit to.
+        return_gateset (set[str]): (optional) The gateset to compile the circuit to.
             e.g. {"cx", "rx",...}. Defaults to the gateset of the target device, or if none is provided, {"cx", "rz", "rx", "ry", "h"}.
         target_device (qiskit.transpiler.Target): (optional) The target device to compile the circuit for. None if no device to target
         custom_passes (list[qiskit.transpiler.TransformationPass]): (optional) A list of custom passes to apply after the default set
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -1,9 +1,8 @@
 from qbraid.programs.alias_manager import get_program_type_alias
 from qbraid.transpiler import ConversionGraph
-from qbraid.transpiler import transpile
-from qiskit.transpiler.passes import BasisTranslator
+from qbraid.transpiler import transpile as translate
 from .transpilers.ucc_defaults import UCCDefault1
-
+from qiskit import transpile as qiskit_transpile
 
 import sys
 import warnings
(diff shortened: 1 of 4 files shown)


ucc/tests/test_compile.py
```diff
@@ -88,47 +88,93 @@ def test_tket_compile():
...
+    """Verify that a custom pass works with a non-qiskit input circuit"""
+
+    class HtoX(TransformationPass):
+        """Toy transformation that converts all H gates to X gates"""
+
+        def run(self, dag):
+            for node in dag.op_nodes():
+                if not isinstance(node.op, HGate):
+                    continue
+                dag.substitute_node(node, XGate())
```
natestemen, line 101: Does this work?
```suggestion
                if isinstance(node.op, HGate):
                    dag.substitute_node(node, XGate())
```

ucc/tests/test_compile.py
```diff
@@ -88,47 +88,93 @@ def test_tket_compile():
...
     analysis_pass = CheckMap(
         t.build_coupling_map(), property_set_field="check_map"
     )
     analysis_pass.run(dag)
     assert analysis_pass.property_set["check_map"]
 
 
-def test_custom_pass():
-    """Verify that a custom pass works with a non-qiskit input circuit"""
+def test_return_gateset():
```
natestemen, line 143: To be consistent with above test:
```suggestion
def test_compile_with_return_gateset():
```

ucc/tests/test_compile.py
```diff
@@ -88,47 +88,93 @@ def test_tket_compile():
...
+    # Check compilation respected the user's gateset
+    dag = circuit_to_dag(result_circuit)
+    gateset = set()
+    for node in dag.op_nodes():
+        gateset.add(node.op.name)
 
-    # Example usage with a cirq circuit, stil showcasing the cross-frontend compatibility
+    # Check that each gate in the compiled gateset is available in the target device
+    for gate in gateset:
+        assert gate in return_gateset
```
natestemen, line 167: Is there a need to go to the DAG? E.g. would something like this work?

```py
assert set(op.name for op in result_circuit).issubset(return_gateset)
```

ucc/tests/test_compile.py
```diff
@@ -147,6 +193,8 @@ def test_compilation_retains_gateset(circuit_function, num_qubits, seed):
     assert analysis_pass.property_set["all_gates_in_basis"]
 
 
+# Skip this test
+@pytest.mark.skip(reason="Slow.")
```
natestemen, line 197: Do we want to disable this?

ucc/tests/test_compile.py
```diff
@@ -88,47 +88,93 @@ def test_tket_compile():
...
-    post_compiler_circuit = compile(cirq_circuit, custom_passes=[HtoX()])
-    assert_same_circuits(post_compiler_circuit, CirqCircuit(X(qubit)))
+# TODO: Write a test to check that the user-defined gateset is available in the target device (funcitonality not yet enforced in ucc.compile)
+# coupling_map = t.build_coupling_map()
+# dag = circuit_to_dag(result_circuit)
+# analysis_pass = CheckMap(
+#     coupling_map, property_set_field="check_map"
+# )
+# analysis_pass.run(dag)
+# assert analysis_pass.property_set["check_map"]
```
natestemen, line 177: Should this be converted to an issue before merging?

ucc/tests/mock_backends.py
```diff
@@ -0,0 +1,68 @@
...
+        self.options.set_validator("shots", (1, 4096))
+        self.options.set_validator("memory", bool)
+
+    @property
+    def target(self):
+        return self._target
+
+    @property
+    def max_circuits(self):
+        return 1024
```
natestemen, line 46: Looks like this function, and the following two functions/methods, are not used. Can we remove them?

ucc/compile.py
```diff
@@ -47,14 +53,33 @@ def compile(
...
+        try:
+            # Use target_device gateset if available
+            return_gateset = target_device.operation_names
+            # Translate into the target device gateset; no optimization
+            compiled_circuit = qiskit_transpile(
+                compiled_circuit,
+                basis_gates=return_gateset,
+                optimization_level=0,
+            )
+        except AttributeError:
```
natestemen, line 80: Is the `AttributeError` from `target_device.operation_names`?

ucc/compile.py
```diff
@@ -37,6 +41,8 @@ def compile(
         return_format (str): The format in which your circuit will be returned.
             e.g., "TKET", "OpenQASM2". Check ``ucc.supported_circuit_formats()``.
             Defaults to the format of the input circuit.
+        return_gateset (set): (optional) The gateset to compile the circuit to.
```
natestemen, line 44:
```suggestion
        return_gateset (set[str]): (optional) The gateset to compile the circuit to.
```

natestemen (commented):
Good stuff!

Can you help me understand why this approach (modifying `ucc.compile`) was preferred over altering `UCCDefault1`? With the current implementation, there are two basis translation passes: one in `UCCDefault1`, and an additional one in `ucc.compile`. Can they be combined?

Part of my hesitation regarding this approach is that there's more complexity in [`qiskit.compiler.transpile`](https://docs.quantum.ibm.com/api/qiskit/compiler#qiskit.compiler.transpile) than there is in [`qiskit.transpiler.passes.BasisTranslator`](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.BasisTranslator). At least that's how I'm seeing it. You have more experience with these than I do, so maybe that's not the case, but I like reaching for the simplest tool.

ucc/tests/test_compile.py
jordandsullivan: This isn't a change from this PR, but yes this works. I can update it.

ucc/compile.py
jordandsullivan: Yes, if a backend doesn't have an attribute defining `.operation_names`

ucc/tests/test_compile.py
jordandsullivan: Ah yep, this does work! Thanks for the suggestion.

ucc/tests/test_compile.py
jordandsullivan: Yeah I will before merging

ucc/tests/test_compile.py
jordandsullivan: Yes, I will do that

ucc/tests/mock_backends.py
jordandsullivan: No, they are required to instantiate the Backend class:
`FAILED ucc/tests/test_compile.py::test_compile_with_target_device - TypeError: Can't instantiate abstract class Mybackend without an implementation for abstract methods '_default_options', 'max_circu...`

jordandsullivan:
> Can you help me understand why this approach (modifying `ucc.compile`) was preferred over altering `UCCDefault1`? With the current implementation, there are two basis translation passes: one in `UCCDefault1`, and an additional one in `ucc.compile`. Can they be combined?
 
Actually I think that's fair, I was trying to avoid adding another argument for `return_gateset` to the `UCCDefault1` signature, but I think it makes sense to do so actually. 

> Part of my hesitation regarding this approach is that there's more complexity in [`qiskit.compiler.transpile`](https://docs.quantum.ibm.com/api/qiskit/compiler#qiskit.compiler.transpile) than there is in [`qiskit.transpiler.passes.BasisTranslator`](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.BasisTranslator). At least that's how I'm seeing it. You have more experience with these than I do, so maybe that's not the case, but I like reaching for the simplest tool.

The reason I used `qiskit.transpile` with no optimization is that if you try to use `BasisTranslator` bare, it doesn't automatically have a mapping for each of the gates that you pass it. You have to provide it an `[EquivalenceLibrary](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.BasisTranslator)` which is a pain, but Qiskit translates it automatically for you.

jordandsullivan:
Code itself is ready for review, but going to wait for #355 to be merged in first before marking Ready for Review.

ucc/tests/__init__.py
```diff
@@ -0,0 +1 @@
+__all__ = ["Mybackend"]
```
natestemen, line 1: What does this line do?

ucc/compile.py
```diff
@@ -47,14 +53,30 @@ def compile(
...
     )
 
+    if target_gateset is not None:
+        # Translate into user-defined gateset; no optimization
+        compiled_circuit = qiskit_transpile(
+            compiled_circuit, basis_gates=target_gateset, optimization_level=0
+        )
+    elif hasattr(target_device, "operation_names"):
+        # Use target_device gateset if available
+        target_gateset = target_device.operation_names
```
natestemen, line 71: I think we should raise an error or warning if a user passed `target_gateset` is different from the `target_device.operation_names`. IMO this could lead to unexpected behavior if a user is not careful. WDYT?

ucc/tests/test_compile.py
```diff
@@ -131,6 +109,52 @@ def run(self, dag):
     assert_same_circuits(post_compiler_circuit, CirqCircuit(X(qubit)))
 
 
+@pytest.mark.parametrize("check", ["opset", "coupling_map"])
+def test_compile_with_target_device(check):
```
natestemen, line 113: I've never seen this pattern before. I would think it would be more natural to have two tests `test_target_device_gateset_respected` and `test_target_device_coupling_map_respected` (modulo naming). Or even just removing the parametrization and the `if`/`elif` blocks and having two asserts. That does technically go against the "only `assert` one thing per test" rule, but IMO the assertion is that the `target_device` is being respected and that takes two assertions.

ucc/tests/mock_backends.py
```diff
@@ -0,0 +1,80 @@
...
+                    UserWarning,
+                    stacklevel=2,
+                )
+        # options = {
+        #     "shots": kwargs.get("shots", self.options.shots),
+        #     "memory": kwargs.get("memory", self.options.shots),
+        # }
+        # job_json = convert_to_wire_format(circuit, options)
+        # job_handle = submit_to_backend(job_jsonb)
+        # return MyJob(self.job_handle, job_json, circuit)
```
natestemen, line 79: Can these be removed?

natestemen (approved):
I like it! A few minor suggestions, but good to go after those are addressed IMO.

jordandsullivan:
Addressed all @natestemen .

ucc/tests/test_compile.py
```diff
@@ -130,6 +108,59 @@ def run(self, dag):
     post_compiler_circuit = compile(cirq_circuit, custom_passes=[HtoX()])
     assert_same_circuits(post_compiler_circuit, CirqCircuit(X(qubit)))
 
+    def test_compile_target_device_opset():
```
bachase, line 111: @jordandsullivan -- I stumbled on this but realized the indenting of this and the next test function make them inner functions inside the `test_custom_pass` function and they are never run. If I restore the formatting properly, the `test_compile_target_device_coupling_map` function is failing on the current version.

There is a potentially another issue with target device still. I'm confirming and will open a consolidated issue for both, but flagging for the moment here.

Issue #342, opened 2025-04-24
Ability to pass gate set as optional argument to `ucc.compile`


natestemen:
**Is your feature request related to a problem? Please describe.**

I'd like an easy way to specify the gate set I'm interested in targeting without the need to create a `target_device` which is very non-obvious how to specify (unless you're very familiar with qiskit).

**Describe the solution you'd like**

```py
ucc.compile(
    circuit,
    gate_set={cirq.X, cirq.Y, cirq.Z, cirq.H, cirq.CNOT},
)
```

**Describe alternatives you've considered**

None

**Additional context**

What happens when a user passes a gate set that is not universal?

jordandsullivan:
So currently the reason we don't do this is that some of the compiler passes assume the specific decomposition at the start of the circuit. 

I think the most straightforward way to address your issue is to decompose again at the end of compilation into your target gateset, which is a feature we can add. But it may not be as efficient as taking the target gateset into account at the beginning of circuit optimization (which would require changes in the logic of our circuit optimization). https://github.com/unitaryfoundation/ucc/blob/main/ucc/transpilers/ucc_defaults.py

jordandsullivan:
Upon further digging, it is actually rather involved to define a [Qiskit target device](https://docs.quantum.ibm.com/api/qiskit/providers#backend-1); there's not a nice way to just say "here's a set of supported gates." Instead you have to manually add them one by one and specify which qubits they are supported on:

```Python
from qiskit.providers import BackendV2 as Backend
from qiskit.transpiler import Target
from qiskit.providers import Options
from qiskit.circuit import Parameter, Measure
from qiskit.circuit.library import PhaseGate, SXGate, UGate, CXGate, IGate
 
 
class Mybackend(Backend):
 
    def __init__(self):
        super().__init__()
 
        # Create Target
        self._target = Target("Target for My Backend")
        # Instead of None for this and below instructions you can define
        # a qiskit.transpiler.InstructionProperties object to define properties
        # for an instruction.
        lam = Parameter("λ")
        p_props = {(qubit,): None for qubit in range(5)}
        self._target.add_instruction(PhaseGate(lam), p_props)
        sx_props = {(qubit,): None for qubit in range(5)}
        self._target.add_instruction(SXGate(), sx_props)
        phi = Parameter("φ")
        theta = Parameter("ϴ")
        u_props = {(qubit,): None for qubit in range(5)}
[...]

PR #355, opened 2025-05-05
338 dynamic circuits qec

jordandsullivan:


jordandsullivan:
My thinking here is that we can adopt the change of removing the `BasisTranslator` steps before and after the main `local_passes` and then in #342 handle the case of user passing target_gates or a device which contains target gates. I will 

- [ ] 1. Add a small test in UCC to make sure we track UCC's ability to handle IF-ELSE statements and control flow. 
- [ ] 2. Remove dev notebooks and scripts from UCC, and
- [ ] 3. Create an issue in ucc-bench to add the qec repetition code as benchmarking circuit

pyproject.toml
```diff
@@ -16,6 +16,7 @@ packages = [{ include = "ucc" }]
 dependencies = [
     "cirq-core>=1.4.1,<2.0.0",
     "ply>=3.11,<4.0.0",
+    "pylatexenc>=2.10",
```
bachase, line 19: Is this new dependency used in the update coded? I'm not finding the spot.

bachase (approved):
👍

Misty-W (approved):
Same comment as Brad, otherwise looks good

pyproject.toml
jordandsullivan: Hm not sure, I think this was an accident. Removed.

Issue #352, opened 2025-04-30
Prototype Fault-Tolerance Checker


bachase:
This is a a parent issue to track progress on prototyping a fault-tolerance checker in UCC. This tracks the phase 1 proposal in #340 , with a goal of recreating the results in "Verifying Fault-Tolerance of Quantum Error Correction Codes" by Chen et al.[^kean] ([paper](https://arxiv.org/abs/2501.14380v2), [code](https://zenodo.org/records/15214439)) but driving the execution from python. From that start point, we would incrementally migrate and generalizes capabilities within UCC.

Issue #340, opened 2025-04-22
Discussion on fault tolerant circuit checker


bachase:
Based on the paper [Verifying Fault-Tolerance of Quantum Error Correction Codes](https://arxiv.org/pdf/2501.14380), create a github discussion outlining a POC implementation, on a small circuit/QEC implementation.

bachase:
Starting proposal in #344

bachase:
Closing given the proposal is now written and now working on it.

PR #356, opened 2025-05-05
Draft: POC fault tolerance checker for cat state preparation circuit

### LINKED ISSUE #353: Enable cat state checking
The most basic code example for the fault-tolerance checker is checking fault tolerance of a cat state preparation circuit. 

Given the code is in Julia, the task is to enable calling the code from python, where the python code can specify the number of faults to check, and return the failure or success of verifying the cat state circuit.

bachase:
Resolves #353 by implementing a fault tolerance checker for the cat state preparation circuit associated with the paper in #352. An example of this working is in [this](https://github.com/unitaryfoundation/ucc/blob/20c2d05ced20f3d3b2bae62af7cb0050fa823630/ucc/tests/test_ft.py#L43) unit test, where a user-provided QASM3 circuit for generating a cat state is verified as FT up to 2 faults, but not up to 3 faults. You can try modifying the circuit, but the code makes a few assumptions still, such as the maximum number of errors to track, that the circuit to check is a QASM3 subroutine with a specific function signature, and it hard codes the target state. All of these can be improved in future versions, but it's likely you'll get errors if you try other circuits.

**To run the unit tests locally**
- You will need to clone this branch and run `uv sync --extras ft` or `uv sync --all-extras` to get the packages needed for the fault tolerance code paths. In the current from, if we end up releasing this, it would only be available to users as an optional install via `pip install ucc[ft]`
- You will also need to install the SMT checker Bitwuzla, following the build instructions [here](https://bitwuzla.github.io/docs/install.html). If we release this, we can consider bundling pre-built version for common platforms

Implementation Details:
* Uses [juliacall](https://juliapy.github.io/PythonCall.jl/dev/juliacall/) and [pyjuliapkg](https://github.com/JuliaPy/pyjuliapkg) to interface with the version of [QuantumSE.jl](https://github.com/njuwfang/QuantumSE.jl) that implements FT checking. This Julia code is in the `ucc/ft/QuantumSE.jl` directory. I've added some functions `PythonTranslate.jl` to simplify the calls across languages.
* The `ucc.ft.checker` module has most of the logic. This includes:
   *  Using `stim.PauliString` and `stim.Tableau` to represent the stabilizers that define the logical qubits. Since this cat state preparation circuit isn't really a QECC, these are just stabilizers that stabilize a cat state.
   * Using a wrapped version of openqasm [printer ast visitor](https://github.com/openqasm/openqasm/blob/e0fc0f135b8ac397a38f41ef01849e98f5d741fb/source/openqasm/openqasm3/printer.py#L125). This is the heavy lifting of translating the QASM3 program to a `@qprog` quantum program that the julia code expects. Lots of assumptions and limitations here since this could be arbitrary code
* For now, the interface for `ucc` users is essentially [`ucc.ft.ft_check`](https://github.com/unitaryfoundation/ucc/blob/20c2d05ced20f3d3b2bae62af7cb0050fa823630/ucc/ft/checker.py#L400), which returns True/False if the circuit is FT (so doesn't give the error that breaks FT yet)

What's Next/What to Review
* Feedback on the interface for users (as considered in the unit test) would be most useful.
* I'd like to add similar checking for one of the error codes covered in the paper/paper's code (surface code?). That would get us to phase 1 of #340 and flush out all the idiosyncrasies of checking a QECC implementation. 
* We could merge this, but it won't be externally useful without a few more iterations. It's also likely we remove the Julia dependency as we generalize and stabilize the interface; hence the thinking to keep as an optional extension for now.


bachase:
CI/CD for unit tests needs to be fixed to handle the new dependencies.

bachase:
Per discussion, closing this and will instead consider housing this prototype in another repo. If/when we decide to integrate into `ucc`, can revisit here.

Issue #377, opened 2025-05-23
Convert the Discussion template to .YAML


jordandsullivan:
Annoyingly, unlike Github Issue templates, you cannot specify Github _Discussion_ templates simply using markdown with a yaml header. Instead you have to specify the whole thing using YAML.

jordandsullivan:
Quite crucial is making sure the label entry in the .yml file perfectly matches the ORGINAL name of the category, as shown in this example repo I just created for testingL

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/ecce291e-7a44-4624-87b1-a13919b89fd3" />

If you change the category name AT ALL, even if you update the label in the yml file to match, it will not pull your template when you create a new discussion using that category. So I created a new Category with the specific name Propose New Compiler Pass and updated the link in https://github.com/unitaryfoundation/ucc/issues/367.

Still to do: transferring the original mardown content fully to yml, which is a bit different bc it has the ability to create forms rather than just pure editable markdown.

jordandsullivan:
I am taking inspiration from [Github's own Community Discussion forms](https://github.com/community/community/blob/main/.github/DISCUSSION_TEMPLATE/accessibility.yml?plain=1)

jordandsullivan:
Frustratingly, it appears that once you Create a Discussion, almost all the elements EXCEPT the user responses get deleted, so you can't go back to edit 😠 . 

As a workaround I am thinking perhaps we should just go back to having it as an issue template rather than a discussion, just due to the extra complexity of the forms.

PR #378, opened 2025-05-23
Convert the .md discussion template to YAML

jordandsullivan:


jordandsullivan:
Annoyingly you can't test the functionality of these discussion templates without merging into main, but I created a [test repo here with an example file](https://github.com/jordandsullivan/ucc-template-test/blob/main/.github/DISCUSSION_TEMPLATE/new-compiler-pass.yml).

and you can test out what the [Propose New Compiler Pass experience](https://github.com/jordandsullivan/ucc-template-test/discussions/new?category=new-compiler-pass) is like.

bachase (approved):
Thanks for setting up the example repo; it made it much easier to test this out.  The changes look fine to me, but per your comments in #377, the workflow will take some getting used to. 

The only alternative I can suggest to reduce the scope of this discussion form to just steps 1 & 2, then use a github issue template for 3-5, and a pull request template for 6 and keep the overall guidelines in the docs/readme. So the user flow would be:

1. Open discussion to propose a pass, get feedback and discuss, share any early results from POCs.
2. Open an issue if you are actively developing and working on it, that has some more detail of the technical strategy and success metrics.
3. Open a PR when ready for review/merging in the pass, where benchmarks will run for validating the success metrics.
4. Overarching - ucc docs and ucc README can summarize this process.

All that being written .. could be overthinking this at this stage.

jordandsullivan:
Yeah I could see splitting it up as you suggest, but I think I prefer a single source of truth that folks can reference at least for now.

jordandsullivan:
We can see how it works for UnitaryHACK this year and make adjustments accordingly.

PR #399, opened 2025-06-02
Reword `ucc.compile()` docstring's `target_gateset` arg

Misty-W:
Update the ucc.compile() docstring to reflect the default behavior when no target gate set is passed: 

> Defaults to the gateset of the target device, or if none is provided, the basis gates of the input circuit are not changed.

User guide already updated in https://github.com/unitaryfoundation/ucc/commit/d259f1749541920e7c72a833abcc76d587872404 - thanks Jordan!


jordandsullivan (approved):
I would just sightly reword and say "Defaults to the gate set of the target device if available. If no target_gateset or target_device is provided, the basis gates of the input circuit are not changed.

PR #401, opened 2025-06-03
MPS encoding

### LINKED ISSUE #381: Implement a new compiler pass (Bounty #3)
_This issue is part of [UnitaryHACK](https://unitaryhack.dev/) 2025, and is 1 of 3 available bounties for implementing and integrating a new quantum compilation pass into UCC._ 

### _What's a compiler pass?_
In UCC (Unitary Compiler Collection), we consider any technique which takes a "raw" uncompiled quantum circuit and returns a circuit which performs better on our [benchmarks](https://github.com/unitaryfoundation/ucc-bench?tab=readme-ov-file#latest-results) to be a good compiler pass. You can work at any layer of the quantum computing stack, but we recommend using what you have easy access to (e.g. low-level hardware access may be difficult on a short time-scale, so perhaps you want to use an emulator) and/ or existing familiarity with (e.g. DAG manipulation). 

## Steps for completing this bounty
If you'd like to pursue this bounty, first be sure to read through this full issue to get a sense of the scope. If you're game, the next step is to create a [New Compiler Pass Proposal](https://github.com/unitaryfoundation/ucc/discussions/new?category=new-compiler-pass) using our Discussions template. Be sure to include the name of your new compiler pass and the bounty # in the title.

The template contains a detailed walk-through of how to implement a new compiler pass into UCC from proposal to development to integration, which you'll update as you go. We will briefly outline the steps here:

### 1. Propose your compiler pass
Tell us about the compiler pass you want to implement, and we'll let you know if it's a good fit for UnitaryHACK.

### 2. Develop your code
Once we've given you the go-ahead, you'll work on a local fork of UCC to develop a prototype and integrate it into UCC.

### 3. Benchmark performance
UCC benchmarks live in the separate [ucc-bench repo](https://github.com/unitaryfoundation/ucc-bench). We have a [suite of quantum circuits](https://github.com/unitaryfoundation/ucc-bench/tree/main/benchmarks) that we regularly benchmark UCC and other popular quantum compilers on. At this stage, you'll benchmark UCC's performance with your new compiler pass integrated. 

---

### Rules
Your compiler pass must improve UCC's performance , which means EITHER:
- Reduce compiled 2-qubit gate counts, decrease runtime, or reduce relative errors on expectation values for our existing [UCC Benchmarks](https://github.com/unitaryfoundation/ucc-bench) 

OR 

- Improve performance on a new benchmark metric you integrate with our help. For instance, we do not currently run quantum hardware benchmarks as part of our standard workflow, so adding a new compiler pass which specifically optimizes based properties of different quantum devices would need a new benchmark.

#### Awarding the bounties
If we receive more than 3 high-quality Pull Requests implementing new passes by the end of the hackathon June 11, we will award the bounties to the contributed PRs which demonstrate the best performance improvement on our [UCC Benchmarks](https://github.com/unitaryfoundation/ucc-bench). 

### Need Ideas?
If you're not sure where to start, check out our [open issues tagged with the `compiler_pass` tag](https://github.com/unitaryfoundation/ucc/issues?q=is%3Aissue%20state%3Aopen%20label%3Acompiler_pass).

ACE07-Sev:
- Closes #381 
- Added `aqc` module.
- Added `mps_sequential.py` which implements Shiju Ran's vanilla encoding.
- Added notebook to demonstrate the performance with avg and worst case.
- Discussed in #394 .


ACE07-Sev:
I didn't add `qmprs` as a .py since I wasn't sure you'd want to adopt it, but I have added it to the notebook to try it out. Just one small note, due to two qubit synthesis having some issues on OS besides Ubuntu, it'd be best to try the code with Ubuntu. I did however make and run the code in windows, so should be fine, but just keep that in mind as a note.

This is short-term, as I have fixed it in `quick`, just not merged with main yet.

ucc/aqc/__init__.py
jordandsullivan, line 1: I'm guessing AQC is approximate quantum compilation?

ucc/aqc/testing mps sequential.ipynb
jordandsullivan, line 1: Thanks for creating this notebook, looks like a good proof of concept. It looks like you're benchmarking against the Qiskit transpiler here, which is fine. The performance of Qiskit is pretty close to UCC, but not exactly the same. You may want to slot in ucc.compile in place of qiskit.transpile to have a fair comparison in this notebook.

ucc/aqc/mps_sequential.py
jordandsullivan, line 1: Could you help me understand how this module will fit in with the existing passes in UCC? It looks like your this MPS encoding operates on a Qiskit quantum circuit object, so I believe it would get called around here, before DAG optimization: 
https://github.com/unitaryfoundation/ucc/blob/main/ucc/compile.py#L58

If that sounds accurate to you, I'd recommend adding a call to the new MPS module there in `ucc.compile` to transform the quantum circuit. This will most likely break some of the tests (since we test for perfect logical equivalence currently: https://github.com/unitaryfoundation/ucc/blob/main/ucc/tests/test_compile.py#L209).

Ultimately if we decided to incorporate this new module, I may recommend adding an optional argument to `ucc.compile` to specify the tolerance within which you'd like the output state to match the approximate encoding. It should default to perfect logical equivalence (i.e. no approximation). For now though, it will probably be easiest to compare against default UCC if you implement the MPS encoding to run every time UCC is called. 

Once you've committed those changes, let us know, and we help you benchmark the performance against default UCC.

jordandsullivan (changes requested):
Looks promising! Next step is to integrate the pass into UCC itself and run our official benchmarks.

ucc/aqc/mps_sequential.py
ACE07-Sev: > It looks like your this MPS encoding operates on a Qiskit quantum circuit object

I can make it work with any package (just used qiskit as an example).

> Could you help me understand how this module will fit in with the existing passes in UCC?

Of course. So, MPS is very good at approximating area-law entangled states with low-rank tensors. Additionally, MPS uses O(N) parameters as opposed to O(2^N), so when you encode the MPS you need O(N) depth. IFF the state you are compiling is a good fit for the MPS, then you will see an exponential reduction in the depth required to compile it with some tiny infidelity. This pass can be used as an unparalleled compression for such states. It's not perfect, as it struggles with volume-law states. I'm working on making a fix to bring its performance on volume-law states to about the same as area-law.

> If that sounds accurate to you, I'd recommend adding a call to the new MPS module there in ucc.compile to transform the quantum circuit. This will most likely break some of the tests (since we test for perfect logical equivalence currently: https://github.com/unitaryfoundation/ucc/blob/main/ucc/tests/test_compile.py#L209).

Yes, I imagine it would break some tests if you check for perfect equivalence.

> Once you've committed those changes, let us know, and we help you benchmark the performance against default UCC.

Copy that.

jordandsullivan:
> Also, do note that `qmprs` will be much more performant in the long-run too, so if you are okay with its dependencies, I would 100 percent recommend that as it's much more powerful, and it's something I am actively working on and based on results we get from here will be updating and modifying it.

@ACE07-Sev if the performance is significantly better by importing qmprs, let's go with that. Do you expect you will continue to maintain qmprs over time as well?

ACE07-Sev:
I hate hyping it, so please try it out through the notebook. If you agree with my statement, then I'll be done with integration in a day.

ACE07-Sev:
Ooh yeah. I'm gonna be working on qmprs for at least the next two years.

jordandsullivan:
@ACE07-Sev Indeed the performance look a lot stronger when using qmprs, esp for volume law circuits. You can go ahead and switch to just importing the library into UCC. We can add it as an optional dependency for now, adding a descriptive message to the doc string for the approximate flag in `ucc.compile` noting that you need to install qmprs to run the approximate synthesis.

image

ACE07-Sev:
That's nice to hear. I'll add it and push ASAP.

ACE07-Sev:
> Since this new pass is your area of expertise, you would need to implement the logic to choose the approximation technique or optimize parameters based on the input circuit, or else just choose a set of parameters that seem decent in most cases.

I was actually working on a smooth parameter defining functionality for this. It'll be there.

> 50 or 100 qubits each in the case of the [compiler benchmarks](https://github.com/unitaryfoundation/ucc-bench/blob/main/benchmarks/compilation_benchmarks.toml)

I need to modify the entrance to the compiler then because if I use statevector it'll be memkilled. I'll figure a way to load directly from circuit instead of circuit -> sv. Thank you for mentioning this.

> Hint: you'll prob want to [clone and install ucc-bench](https://github.com/unitaryfoundation/ucc-bench/tree/main?tab=readme-ov-file#running-or-development) and use the instructions here to [install your dev fork of ucc](https://github.com/unitaryfoundation/ucc-bench/tree/main?tab=readme-ov-file#upgrading-ucc-version) within your ucc-bench environment. From there, you can [run benchmarks](https://github.com/unitaryfoundation/ucc-bench/tree/main?tab=readme-ov-file#usage-running-a-benchmark-suite) locally to test out different parameters for your optimization.

Copy. I'll finish the mods, and then I'll get to this one. I have to do a bit of testing to fine-tune the parameters. I'm gonna go sleep now (4 am here, so I'm a bit useless after this hour hehe), and hopefully get these done by end of tomorrow.

ACE07-Sev:
Thank you! So, I am pretty much finished with the code, just need to add in the direct loading from circuit to avoid memory issues, and a tiny bit more playing with it to fine-tune it for maximum fidelity and minimum depth, and I should be done.

jordandsullivan:
@ACE07-Sev Cool! Looking forward to reviewing.

ACE07-Sev:
Apologies for the delay, just working a bit more on improving fidelity for general case, and specifically for volume-law as much as possible.

ACE07-Sev:
Some notes after performing further experiments and doing a bit more research on the theory side:
- `ucc.aqc.QmprsCompiler()` should be used mainly for *area-law* entangled circuits to provide significant compression whilst maintaining high fidelity. Depending on the use-case, we can modify the automated parameter generation to optimize the performance (tailored automation vs generic automation), or sacrifice some runtime slightly to perform more sweeps.
- `ucc.aqc.QmprsCompiler()` and any and all TN compilers do not by definition perform well with *volume-law* circuits (hence why we use quantum computing in the first place, otherwise we would be just using TN simulators on GPU clusters). That being said, depending on the entanglement entropy we can still use `ucc.aqc.QmprsCompiler()` to get significant depth reduction BUT at the cost of considerable infidelity (again depending on the circuit). This however **won't** make `ucc.aqc.QmprsCompiler()` obsolete for such cases, but instead provides its benefit as an initial good guess and can then be used in a QML pipeline to be fine-tuned to match the target fidelity (I can add this if my laptop cooperates, otherwise I can do it after the event via `qmprs` to lessen the code on your side). That way, you still get the compression, but only suffer slight runtime to match the fidelity of area-law to some extent.

So, the `ucc.aqc` module in short will be a complement to the `ucc.compile` where it compresses circuits coming its way **where possible** (and use qml for volume-law circuit to provide as consistent of a behavior as mathematically possible) and if worst case comes and it fails, it will simply be ignored which makes it at worst what `ucc.compile` already does, and on avg much better for certain cases.

ACE07-Sev:
The only thing I haven't added yet is direct conversion of circuits to MPS to avoid getting memkilled. This again depends on the circuit type, as volume-law states require exponentially large bond dimension and thus require more memory. I can add something efficient for area-law which would considerably mitigate memory issues when loading the circuit.

jordandsullivan:
Thanks @ACE07-Sev . Can you try to run the official ucc-benchmarks with your new pass included? I would recommend running with both the area law and the volume law with some sensible defaults for parameters (or a few sweeps).

ACE07-Sev:
Copy copy.

ACE07-Sev:
@jordandsullivan Do you care about GHZ circuits or should I skip them?

jordandsullivan:
Thanks @ACE07-Sev . I think that next next step will be for you take your benchmarks from the notebook and implement them in ucc-bench, making sure you are able to measure the performance with appropriate parameters. 

These two instructions will be your friends here: 
1. Using the benchmarks on a local fork of UCC: https://github.com/unitaryfoundation/ucc-bench?tab=readme-ov-file#upgrading-ucc-version
2. Adding a new circuit or metric to ucc-bench: https://github.com/unitaryfoundation/ucc-bench?tab=readme-ov-file#updating-a-benchmark-eg-adding-a-new-circuit-or-metric

ACE07-Sev:
Copy that. Could I ask what are your current thoughts on the benchmark results?

ACE07-Sev:
Have you had a chance to view benchmark.ipynb? My request was wrt that.

bachase:
@ACE07-Sev thanks for all this consideration and progress so far. With regards to `ucc-bench` and benchmarks, my view:

1. I understand the challenge in running on larger circuits. There are smaller circuits in this [benchmark](https://github.com/unitaryfoundation/ucc-bench/blob/main/benchmarks/simulation_benchmarks.toml). I believe this overlaps somewhat with the set of generated circuits you have in the notebook. 
2. What we are interested in knowing is how this new pass performs relative to what the existing baseline `ucc` performs on those circuits. I didn't see that comparison in the notebook. Is there a way you can add it? If not, I can try to help do a manual run benchmark run against the code, but it would need your branch to pass unit tests first. The motivation here is to establish what gains this method has over the baseline, and at what cost in fidelity.
3. There's a lot of (good!) detail on the area vs volume law regimes, but that is making it harder to understand the impact of this approach for a user of `ucc`. That is, we wouldn't expect user to be an expert in this technique or which circuits to try it on. Ideally, the code is able to determine when to try this approach and apply it in just those circumstances. 

As an aside from 3 and 4, the more focused the notebook/comments, the easier it will be for UCC maintainers to review.

ACE07-Sev:
Thank you very much for the kind words.

> I understand the challenge in running on larger circuits. There are smaller circuits in this [benchmark](https://github.com/unitaryfoundation/ucc-bench/blob/main/benchmarks/simulation_benchmarks.toml). I believe this overlaps somewhat with the set of generated circuits you have in the notebook.

I'm currently adding some memory guardrails to not error out or get memkilled so that when it can't allocate memory, it'd fall back on your default compiler.

> What we are interested in knowing is how this new pass performs relative to what the existing baseline ucc performs on those circuits. I didn't see that comparison in the notebook. Is there a way you can add it? If not, I can try to help do a manual run benchmark run against the code, but it would need your branch to pass unit tests first. The motivation here is to establish what gains this method has over the baseline, and at what cost in fidelity.

Of course. So, in the `benchmark.ipynb` I do exactly that for `qcnn`, `ghz`, `trivial_bvlike`, and `random_clifford` circuits. I currently only added the `random_area_law` which is the best instance of what `qmprs` should be useful. If you have a look at `benchmark` function in that notebook, you can see what I'm comparing against, which is your `compile` from `compile.py`.

> There's a lot of (good!) detail on the area vs volume law regimes, but that is making it harder to understand the impact of this approach for a user of ucc. That is, we wouldn't expect user to be an expert in this technique or which circuits to try it on. Ideally, the code is able to determine when to try this approach and apply it in just those circumstances.

Of course, those notebooks are primarily intended for making your work easier. `ucc` users only need to know that if they set `aqc` to `True`, then whenever the circuits they are compiling fall under area-law they can expect a significant reduction in depth with 0.87+ fidelity on avg with the current fine-tuning (which with a feature I'm working on `qmprs` should be slightly improved to hopefully 0.90+). The edge cases where area-law would see worse performance is when their circuits are very trivial, to the point that even a single layer is too much (which they get 1.0 fidelity, with slightly less cx count but possibly more depth as can be seen in `qcnn` case).

Point is that when the circuits they want to compile are a good fit for MPS, then they can expect considerable reduction in CX count and depth at the cost of slight infidelity. For instance, say you want to compile a state preparation circuit to optimize its depth. This is used in many algorithms. Traditionally, you're stuck with $O(2^N)$. MPS brings that to approximately $O(N)$. The fidelity will be consistently above 0.87 assuming area-law circuits. That is a huge advantage, and why we use MPS in the first place.

> Ideally, the code is able to determine when to try this approach and apply it in just those circumstances.

Exactly. That's why I added entanglement calculators to automatically decide parameters for optimal fidelity and depth. This removes the need for having the know-how. Right now, the code automatically decides the parameters. I am confident I can improve this but it realistically needs a month of benchmarking as each benchmark can take ~1-2 hours. Not a time we have right now wrt the event, but something I will 100 percent do.

Lastly, if the fidelity of the compiled circuit is poor, then it'll just be ignored which makes the performance as good or better than what is already there. I can add in more strict fidelity requirements if you like as well, but that's the overall idea of it and why it's a good fit here.

"When it can guarantee adequate fidelity, and the depth is lower than before, it'll apply compilation. Otherwise, it'll be discarded and everything will be same as your baseline."

ACE07-Sev:
Sure, I'll get unit-tests working. I was working on the items Brad mentioned, and will push once I wake up.

ACE07-Sev:
I ran the tests locally. Will make sure to get it compatible with qmprs missing as well.
```
(.venv) @ACE07-Sev ➜ /workspaces/ucc (main) $ pytest ucc/tests
=================================================================================================== test session starts ===================================================================================================
platform linux -- Python 3.12.3, pytest-8.4.0, pluggy-1.6.0
rootdir: /workspaces/ucc
configfile: pyproject.toml
collected 94 items                                                                                                                                                                                                        

ucc/tests/test_compile.py .............................................................................................. 
```
All pass.

jordandsullivan:
> So, I have a question. Right now I have updated the compiler to do the following:
> 
> ```
> if aqc is True:
>     compiled_circuit = aqc compiled
>     if fidelity <= 0.8:
>         discard compiled circuit and return original circuit
>     endif
>     
>     if compiled depth >= qiskit transpiled depth of original AND compiled cx count >= qiskit transpiled cx count of original:
>         discard compiled circuit and return original circuit
>     endif
>     
>     return compiled circuit by aqc
> endif
> ```
> 
> So it will guaranteed be better than or equal to yours with guaranteed minimum fidelity permitted by you (re-ran benchmarks and confirmed it works).
> 
> Now, I would like to know if you want these exact conditions, or something different. I want to try to benchmark as much as I can to find the optimal parameter generation before pushing and getting it run by your ucc-bench CI.

Those conditions seem reasonable to me. Depending on how long the runtime is, we may want to simplify the logic and just use a heuristic, but for now, let's just use what you have. It doesn't need to be optimal or perfect; let's just make sure it works with the rest of the compilation pipeline without any errors :) That's why I asked if you could fix errors coming up in the CI/CD tests.

ACE07-Sev:
I'll fix the formatting now.

ucc/aqc/benchmark new approach.ipynb
bachase, line 1: Just confirming - is this showing the same performance for the two compilers for GHZ? 
image

bachase: Also, although these notebooks are useful for this review, it's a bit out of place to have them all included in the ucc package. I do see a benefit of some context for this method, so recommend either a shorter readme or a single notebook, that is much less verbose and focused on the method and when it is used.

To be a tad blunt -- the volume of detail in these notebooks made it hard to understand the approach and results. I appreciate all the work, but it makes it harder to follow. I'd go with the less is more approach here.

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "qbraid>=0.9.3,<1.0.0",
     "qiskit>=1.4.2,<3.0.0",
     "qiskit-qasm3-import>=0.5.1,<1.0.0",
+    "quimb==1.10.0",
```
bachase, line 23: Please run `uv lock` and commit the updated lock file.

ucc/compile.py
```diff
@@ -24,13 +29,115 @@
     )
 supported_circuit_formats = ConversionGraph().nodes()
 
+# Check if `qmprs` is installed for AQC compilation
+qmprs_available = importlib.util.find_spec("qmprs") is not None
+
+
+def has_enough_memory(num_qubits: int) -> tuple[bool, float, float]:
```
bachase, line 36: Can you move this to the aqc module, since it is specific to that pass?

ucc/compile.py
```diff
@@ -24,13 +29,115 @@
...
+    # requires at worst-case (volume-law)
+    # statevectors use np.complex128 which needs 16 bytes
+    # Use half of the memory available to store the IR
+    memory_required_gb = 2 ** (4 + num_qubits - 31)
+    has_memory = memory_required_gb <= available_memory_gb
+
+    return has_memory, memory_required_gb, available_memory_gb
+
+
+def approx_compile(circuit):
```
bachase, line 63: Could this be part of the aqc pass itself, rather than a pre-pass phase?

ucc/compile.py
bachase: Alternatively you could just inline this to the function below.

ucc/compile.py
```diff
@@ -56,12 +165,17 @@ def compile(
 
     # Translate to Qiskit Circuit object
     qiskit_circuit = translate(circuit, "qiskit")
+
+    # Utilize approximate quantum compilation via MPS encoding
```
bachase, line 169: As an alternative implementation, would it make more sense for aqc to be a custom pass? Given the user has to opt-in, that seems to match other patterns and also the approach for another optional method in #412.

ucc/compile.py
```diff
@@ -24,13 +29,115 @@
     )
 supported_circuit_formats = ConversionGraph().nodes()
 
+# Check if `qmprs` is installed for AQC compilation
+qmprs_available = importlib.util.find_spec("qmprs") is not None
```
bachase, line 33: Since this is only used once, I'd either inline this check to the function below, or perhaps better, import this from the aqc module so its defined in one place.

ucc/aqc/qmprs_compiler.py
```diff
@@ -0,0 +1,98 @@
...
+        circuit = self.sequential.prepare_state(
+            statevector=statevector,
+            bond_dimension=2**num_qubits,
+            num_layers=num_layers,
+            num_sweeps=num_sweeps,
+        )
+
+        if verbose:
+            fidelity = np.vdot(circuit.get_statevector(), statevector)
+            print(
```
bachase, line 91: Rather than bare prints/verbose, I'd recommend using [the python logger](https://docs.python.org/3/library/logging.html) to have this output with more composability.

ucc/aqc/mps_sequential.py
```diff
@@ -0,0 +1,326 @@
+import numpy as np
+from numpy.typing import NDArray
+import quimb.tensor as qtn  # type: ignore
+from qiskit import QuantumCircuit  # type: ignore
+from qiskit.quantum_info import Statevector  # type: ignore
+from .utils import calculate_entanglement_entropy_slope
+import warnings
+
+
+def gram_schmidt(matrix: NDArray[np.complex128]) -> NDArray[np.complex128]:
```
bachase, line 10: (asking to understand) is there a reason to implement this yourself versus using scipy.linalag.qr and taking the columns in Q?

ucc/aqc/mps_sequential.py
```diff
@@ -0,0 +1,326 @@
+import numpy as np
+from numpy.typing import NDArray
+import quimb.tensor as qtn  # type: ignore
```
bachase, line 3: My reading was this module should be used if quimb package is not available. So would this line error out if that were the case?

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
...
 ]
```
bachase, line 25: Should qmprs be added as an optional/dev dependency? If not (which is fine), you should at least add a README detail on how users would need to install it to leverage that capability.

ucc/compile.py
```diff
@@ -56,12 +165,17 @@ def compile(
 
     # Translate to Qiskit Circuit object
     qiskit_circuit = translate(circuit, "qiskit")
+
+    # Utilize approximate quantum compilation via MPS encoding
+    # This is useful for area-law entangled circuits which can
```
bachase, line 170: Are there other restrictions on the circuit to check, e.g. it has mid-circuit measurement, or other properties that won't work with this method?

bachase (commented):
Thanks for getting this in good shape! I left some comments+feedback, but I think the highest design one to decide is whether this should be an optional pass instead of behind this flag. 

I'll next look to run the code and give any feedback on it's performance.

pyproject.toml
ACE07-Sev: Copy that.

ucc/compile.py
ACE07-Sev: Sure, will move it to aqc. Although, useful for any compiler that requires considerable memory.

ucc/aqc/qmprs_compiler.py
ACE07-Sev: Copy that.

ucc/aqc/mps_sequential.py
ACE07-Sev: No, this is used if `qmprs` is not available. Without `quimb` we can't make MPS. I rely on `quimb` for handling the low-level tensor operations.

pyproject.toml
ACE07-Sev: Yes I'll add it. I couldn't there because at the moment I haven't made a pypi publish for it given I'm finishing a few changes for it and `quick`.

ucc/compile.py
ACE07-Sev: I don't think it would work with mid-circuit measurement purely because measurement operator is non-unitary (qiskit will scream because it can't define a statevector from the circuit with measurements in it). The circuit that we pass to optimize has to be unitary (this is a condition for most circuit optimizers and statevector compilers, where MPS is both in this case).

ucc/aqc/mps_sequential.py
bachase: Ah my mistake. I tranposed those package names!

ucc/aqc/benchmark new approach.ipynb
ACE07-Sev: > Just confirming - is this showing the same performance for the two compilers for GHZ?

Yes. If you can't beat 'em, join 'em. This is the fallback protocol for when the compiled circuit by aqc is more expensive than the original (which in this case it is since GHZ is a really shallow circuit).

ACE07-Sev:
@bachase It should now be able to run without error sir even in absence of qmprs. I re-ran benchmarks and assured the difference is as minimal as possible (albeit without suffering more depth given the magic of sweeping :) ).

bachase:
@ACE07-Sev I updated to the latest version of the code over in https://github.com/unitaryfoundation/ucc-bench/pull/85 and got the results below.  The benchmarks don't check the logical equivalence of pre-post compilation circuits, or how close they are.

```csv
compiler,benchmark_id,target_device_id,raw_multiq_gates,compile_time_ms,compiled_multiq_gates
ucc,qaoa,,96,16.9898,96
ucc,qv,,150,6.3619,150
ucc,qft,,105,4.5471,105
ucc,square_heisenberg,,144,2.7142,36
ucc,prep_select,,658,27.7293,948
ucc,qcnn,,32,8.3921,32
ucc+mps,qaoa,,96,2894.3682,96
ucc+mps,qv,,150,2851.6386,150
ucc+mps,qft,,105,2399.9511,0
ucc+mps,square_heisenberg,,144,2393.3488,0
ucc+mps,prep_select,,658,416.7532,27
ucc+mps,qcnn,,32,383.3896,27
```

Is it surprising that say the qft circuit is compiled down to a circuit with no multi-qubit gates? https://github.com/unitaryfoundation/ucc-bench/blob/main/benchmarks/circuits/benchpress/qft_N010_basis_rz_rx_ry_cx.qasm is the circuit that is compiled.

ACE07-Sev:
I'll test it. It is surprising unless the CX gates completely cancel out.

ACE07-Sev:
Here's the entanglement entropy plot for the state.
image

You see how it's practically zero entanglement? That's why it's compiling it perfectly with no CX gates.

ACE07-Sev:
My guess:
```
ucc+mps,qaoa,,96,2894.3682,96 # It's just falling back to ucc with aqc=False
ucc+mps,qv,,150,2851.6386,150 # it's just falling back to ucc with aqc=False
ucc+mps,qft,,105,2399.9511,0 # trivial case where entanglement is practically 0
ucc+mps,square_heisenberg,,144,2393.3488,0 # trivial case where entanglement is practically 0
ucc+mps,prep_select,,658,416.7532,27 # Actually where it's supposed to be used, this is what I like
ucc+mps,qcnn,,32,383.3896,27 # Slight reduction in cx count, but I imagine more circuit depth
```

bachase:
As an aside, I added an operator based check as part of the benchmarking branch https://github.com/unitaryfoundation/ucc-bench/blob/7084c157bdb5b2e6c94f581435a6edc87d2a3b75/src/ucc_bench/utils.py#L5-L9

```
def validate_circuit_equiv(input, output):
    input_qc = transpile(input, "qiskit")
    output_qc = transpile(output, "qiskit")

    return Operator(input_qc) == Operator(output_qc)
```

The existing ucc transpiler passes this check, but this fails for version that enables the MPS based pass for e.g. qft. I know we don't expect full equivalence; is there another operator distance to consider that would be relevant here?

ACE07-Sev:
> But for compilation purposes, I don't think ucc wants to assume that generally, as users might be combining the post-compiled circuits in a variety of ways.

Of course. That's what MPO encoding is for (something I'm working to add soon). That takes a unitary to compile as opposed to a statevector, and that given it's implementing an operator would be independent of what state you have before applying its circuit.

ACE07-Sev:
> The existing ucc transpiler passes this check, but this fails for version that enables the MPS based pass for e.g. qft. I know we don't expect full equivalence; is there another operator distance to consider that would be relevant here?

Yes, that's fully expected given the qft you convert to circuit has CX gates, and a ton other gates. What MPS does is find the optimal circuit in its own way that would give you the same state. For MPS you need to evaluate based on state equivalency/distance, not operator.

For MPO later on when I add it then you have to evaluate based on operator.

ACE07-Sev:
Anything I need to add/change?

bachase:
Thanks for the continued work on this @ACE07-Sev. Those updated benchmarks helped me better understand how MPS approach is working and when its beneficial to use. I think including this as an optional pass is a good near-term add to UCC, and I'm excited to see the MPO approach for wider applicability longer term.

To merge this contribution, I'd request
* Removing the ipynb notebooks and adding a briefer README.md on the method within the aqc directory
* If possible (happy to help lend a hend too), I'd prefer this implemented as a custom pass for now, similar to the approach in #412. That better fits the paradigm UCC had in mind for optional compilation phases. If there's a reason not to do this though, please share.
* Update the [user guide](https://github.com/unitaryfoundation/ucc/blob/main/docs/source/user_guide.rst) briefly explaining how to enable this as an optional pre-compile pass. Again, can compare with how this is done in #412.

The only mystery that remains to me is why the "official" benchmarks are [changing](https://github.com/unitaryfoundation/ucc/pull/401#issuecomment-2963565628); by default this pass is off so I wouldn't expect anything. I'd like to have an explanation for that before merging, but don't have any good leads to investigate at the moment.

ACE07-Sev:
> Removing the ipynb notebooks and adding a briefer README.md on the method within the aqc directory

Copy that.

> If possible (happy to help lend a hend too), I'd prefer this implemented as a custom pass for now, similar to the approach in https://github.com/unitaryfoundation/ucc/pull/412. That better fits the paradigm UCC had in mind for optional compilation phases. If there's a reason not to do this though, please share.

Sure, I'll see how it's done and will amend.

> Update the [user guide](https://github.com/unitaryfoundation/ucc/blob/main/docs/source/user_guide.rst) briefly explaining how to enable this as an optional pre-compile pass. Again, can compare with how this is done in https://github.com/unitaryfoundation/ucc/pull/412.

Sure I'll add it in.

> I'd like to have an explanation for that before merging, but don't have any good leads to investigate at the moment.

I'll go through the changes and see if I can find it.

bachase:
Sounds good @ACE07-Sev . If you focus on the first 3 first, we can wait and rerun the benchmarks after.

ACE07-Sev:
> If you focus on the first 3 first, we can wait and rerun the benchmarks after.

Sure, I'll finish the first 3 items.

bachase:
> When you say custom pass, you mean add it under `transpilers` module?

Yes, you could do it in `transpilers.aqc`, and then wrap your existing code in a new pass like in the [docs](https://ucc.readthedocs.io/en/latest/user_guide.html#writing-a-custom-pass). So like:

```
# in a file transpilers/aqc/pass.py
from qiskit.transpiler.basepasses import TransformationPass
from qiskit.dagcircuit import DAGCircuit
from . import whatever_you_need

class MyCustomPass(TransformationPass):

    def __init__(self):
        super().__init__()


    def run(self, dag: DAGCircuit) -> DAGCircuit:
        #  Your code here .. basically the current approx_compile?
        return dag
```

ACE07-Sev:
Understood. Thank you.

ACE07-Sev:
@bachase Sorry to bother. Is the qft QASM too long for `user_guide.rst` example?

ACE07-Sev:
> brief code snippet for how to add it as a custom pass or configure it.

Yes, I followed the example you pointed. I see, so I guess just sth to show the syntax. I'll use sth smaller.

ACE07-Sev:
Sorry for the delay. Trying to get some latex working in `README.md` for `ucc/transpilers/aqc`.

ACE07-Sev:
I'll fix the lint issue by importing directly then.

docs/source/user_guide.rst
```diff
@@ -155,6 +155,42 @@ Alternatively, we can add our custom pass, as shown in the following example.
       circuit_to_compile, custom_passes=[MyCustomPass()]
    )
 
+An example of a custom pass: Approximate Quantum Compilation via MPS encoding
+=============================================================================
+The ``MPSEncoder`` is a custom pass provided in ``ucc.aqc``. Users can opt for `qmprs<https://github.com/Qualition/qmprs>`_ for a more advanced implementation of the same pass.
```
bachase, line 160: By opt for, do you mean they would need to also install that package? If so, please be explicit about that instruction and how to do it.

docs/source/user_guide.rst
```diff
@@ -155,6 +155,42 @@ Alternatively, we can add our custom pass, as shown in the following example.
...
+This pass leverages Matrix Product State (MPS) representation of a state to approximately compile the state to a quantum circuit using multiple layers of one and two qubit gates in O(N) depth.
+The automatic parameter definition takes the entanglement structure of the input state into account, and tries to come up with the optimal parameters to maximize fidelity and minimize circuit depth. Users can also override ``optimal_params`` static method to define their own rule for generating the optimal parameters.
```
bachase, line 162: I would add something to help users know about the input state assumption, so something like

> Most quantum circuit libraries are written wassuming the initial state is all zeros in the computational basis. This pass's optimization may rely on that assumption. If you intend to run your post-compiled circuit on other input states, or in sequence with other circuits, be aware that this pass might not be equivalent in those cases.

ucc/transpilers/aqc/README.md
```diff
@@ -0,0 +1,39 @@
...
+unitaries. What Ran approach proposes instead is to truncate the MPS bond dimension to 2, thus getting only one and two qubit gates, and to accommodate the fidelity issue
+by using multiple layers ([[2]] Figure 1).
+
+Ran's approach iteratively generates unitary layers that gradually disentangle the target state, and once disentangled enough it applies the layers in reverse order to prepare
+the target state ([[2]] Section A, Algorithm 1).
+
+### Qmprs Compiler
+
+`qmprs` [[3]] is a state-of-the-art approximate quantum compiler which provides approximate quantum compilation via MPS/MPO representations. Compared to Ran's approach which is implemented in `mps_sequential.py`, `qmprs` provides higher fidelity truncations, as well as utilization of environment tensor updates to variationally optimize the layers generated by Ran to improve fidelity without changing circuit depth [[2]]. This provides consistently better fidelity which means less depth to achieve similar fidelities.
+
```
bachase, line 36: Consider adding a line here like I suggested in the docs to let users know that the current MPS version is a *state* compiler and not a general *operator* compiler. 

To help you understand why I keep coming back to that, `ucc` is a general purpose quantum compiler for arbitrary circuits/algorithms, whereas this specific technique is more focused on states. I'd just want to ensure users clearly know that distinction and when to pick this pass.

ucc/transpilers/aqc/mps_sequential.py
```diff
@@ -0,0 +1,323 @@
...
+                )
+            )
+
+            fidelity = np.vdot(disentangled_mps.to_dense(), zero_state)
+
+            if fidelity >= self.max_fidelity_threshold:
+                # If the disentangled MPS is close enough to the zero state,
+                # we can stop the disentanglement process
+                if verbose:
+                    print(
```
bachase, line 229: Similar to the other feedback you already addressed, please switch all bare prints throughout the module to `logging` calls. You can then remove the verbose flag, as the logging configuration will manage the output.

bachase (commented):
Looking close!  

I'm still perplexed on why the benchmarks are still showing a change in timing with this

docs/source/user_guide.rst
```diff
@@ -155,6 +155,46 @@ Alternatively, we can add our custom pass, as shown in the following example.
       circuit_to_compile, custom_passes=[MyCustomPass()]
    )
 
+An example of a custom pass: Approximate Quantum Compilation via MPS encoding
+=============================================================================
+The ``MPSEncoder`` is a custom pass provided in ``ucc.aqc``. Users can opt for `qmprs<https://github.com/Qualition/qmprs>`_ for a more advanced implementation of the same pass.
```
bachase, line 160: Note the build is failing because this link is missing a space between `qmprs` and the `<htt

bachase:
Closing in favor of #421

PR #402, opened 2025-06-03
Allow triggering benchmark comparision on fork PRs

bachase:
This is a companion to https://github.com/unitaryfoundation/ucc-bench/pull/69, and switches the workflow trigger to `pull_request_target` for the benchmark PR comparison runs. This should enable PRs with the `preview-benchmark-results` label that are coming from forks to run the benchmarks.


jordandsullivan (approved):
👍🏼 Looking forward to this one

PR #405, opened 2025-06-05
Add section to docs on research using UCC

### LINKED ISSUE #388: Highlighting uses of UCC
This task is to add a way to higlight uses of UCC, motivated by the recent paper https://arxiv.org/pdf/2505.09036.  For now, this could be a page within the docs.

bachase:
Resolves #388 by adding a dedicated page to the docs to summarize and link out to uses of UCC. The flow is to add a bibtex entry into a new `refs.bib` file, then add a summary which cites that in the documentation file. It's a little overboard for now, but felt it would be better over time to make formatting consistent and to have a portable set of references.

I also added an issue template to walk users through doing this in the future.

### DIFF
--- a/docs/source/research_references.rst
+++ b/docs/source/research_references.rst
@@ -1,0 +1,14 @@
+Research References
+======================
+
+This page lists research papers that have used the Unitary Computing Compiler (UCC) in their work.
+If you have used UCC in your research, please let us know by `opening an issue or pull request <https://github.com/unitaryfoundation/ucc/issues/new?template=research_reference.md>`_ on GitHub to update this list.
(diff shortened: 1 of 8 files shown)


jordandsullivan (approved):
Cool!

docs/source/research_references.rst
```diff
@@ -0,0 +1,14 @@
+Research References
+======================
+
+This page lists research papers that have used the Unitary Computing Compiler (UCC) in their work.
+If you have used UCC in your research, please let us know by opening an issue or pull request on GitHub to update this list.
```
natestemen, line 5: You should be able to add a link to the new issue template using syntax like this:

> https://github.com/unitaryfoundation/ucc/issues/new?template=bug_report.md

natestemen (approved):

PR #431, opened 2025-06-18
Add Intro video about UCC

WingCode:
While doing UnitaryHack 2025 and coming across UCC as a newbie, this video helped me a lot to understand the overview of the project. So, I think it would be helpful for other new comers to the project too!


jordandsullivan (approved):
I like it @WingCode ! Glad you found this video helpful :)

PR #442, opened 2025-06-24
Ensure all pre-commit hook are validated in CI

### LINKED ISSUE #413: Ensure all pre-commit hooks are validated in CI (or disable)
As noted here https://github.com/unitaryfoundation/ucc/pull/412#discussion_r2140230170_, our CI is only checking the ruff linting checks and not the other pre-commit hooks. We should either (1) enforce those in CI as well or (2) disable if we don't find them useful.

bachase:
Resolves #413

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,12 @@ jobs:
       - name: Install the project
         run: uv sync --all-extras --all-groups
 
+      - name: Install pre-commit
+        run: uv pip install pre-commit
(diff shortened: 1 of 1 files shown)


bachase:
https://github.com/unitaryfoundation/ucc/actions/runs/15859361551/job/44712465530?pr=442 shows this ran successfully.

jordandsullivan (approved):
LGTM thanks @bachase !

PR #455, opened 2025-07-17
Fix links to research references in docs

jordandsullivan:
The previous links were missing a ../ in the relative path


bachase (approved):
:+1:

Issue #341, opened 2025-04-24
Add default gate set to docs


natestemen:
I'm trying to understand what gate set running `ucc.compile` compiles to, but I cannot find it! Not exactly sure where it should go, but a helpful piece of info to display.

bachase:
The default basis is here https://github.com/unitaryfoundation/ucc/blob/9c8430e0b6eb63d280d30539b76db1bfde44fda2/ucc/transpilers/ucc_defaults.py#L52.

Agree it should be added to the docs if not there already.

jordandsullivan:
@bachase Shall we remove this one from the milestone? per #411

PR #492, opened 2025-08-21
Simplified and optimized logic for target gateset selection. 

jordandsullivan:
The current operation is 
1. Identify the target gateset:
    a) `target_device.operation_names` if a target_device is provided, or
     b) the target_gateset if that is provided, or
     c) else {cx, rz, rx, ry, h} if neither is provided.
2. Translate the initial circuit into the provided gateset first, pre-optimization (This may not be necessary, as commenting it out doesn't break any tests, so UnitarySynthesis may be taking care of this).
3. Pass the target gateset to the UnitarySynthesis pass to optimize for that target basis. 

1c) is different than the previous behavior of leaving circuits alone if no target was specified. This way we aren't losing out on the optimization potential of the UnitarySynthesis pass.


bachase (commented):
👍 this is much clearer.

Approving, but would be great to see the benchmark suite runs to confirm the expected impact. (Pending https://github.com/unitaryfoundation/ucc-bench/pull/126)

This might also address #341 ?

Issue #411, opened 2025-06-10
Consider mix of `target_device` and `target_gateset`


bachase:
In adding layout benchmarks to [ucc-bench
](https://github.com/unitaryfoundation/ucc-bench/pull/78), we noticed that qiskit and UCC differ when the user passes both a  `target_gateset`/`basis_gates` and `target_device`.

From Qiskit docs

> The prioritization of transpilation target constraints works as follows: if a ``target``
    input is provided, it will take priority over any ``backend`` input or loose constraints
    (``basis_gates``, ``coupling_map``, or ``dt``). If a ``backend`` is provided
    together with any loose constraint

So for this case, the gateset in `target` overrides the `basis_gates` passed to `qiskit.compiler.transpile`.

In `ucc`, we convert to `target_gateset` at the end, regardless of whether `target` provides its own gatesets. 

It's worth considering how these options interact for users coming from qiskit.

bachase:
@jordandsullivan when looking into #341, I took a closer look at this code. The documentation states if neither `target_gateset` or `target_device` is specified, then the circuit's basis gateset should be unchanged. However, `UCCDefault` *does* specify a specific default gateset [here](https://github.com/unitaryfoundation/ucc/blob/583179f5d051537b30aadd1efff96cae533da247/ucc/transpilers/ucc_defaults.py#L52) that is used by one pass [here](https://github.com/unitaryfoundation/ucc/blob/583179f5d051537b30aadd1efff96cae533da247/ucc/transpilers/ucc_defaults.py#L78). As a result, the test below will fail as the basis is changed.


```python
def test_compile_with_no_target_gateset_or_device():
    """Test that the final circuit does not change the gateset if no target gateset or device is provided"""
    circuit = QiskitCircuit(2)
    circuit.cx(0, 1)
    circuit.h(0)

    result_circuit = compile(
        circuit,
    )

    assert set(op.name for op in result_circuit).issubset({"cx", "h"})
```

Is this an error? I also wonder if we should remove the default gateset in `UCCDefault` in favor of the approach in `compile`?

jordandsullivan:
Ah good point. I am open to removing the default gate set in UCCDefault1, but I would want to understand how that affects circuits which are not already decomposed into 2 qubit gates, which all.of our benchmarks currently are.

bachase:
[...]

PR #497, opened 2025-08-28
Draft: Fix use of target device

bachase:
This is a partial fix for an issue introduced in #354 in which some unit tests where indented and not being run. This was hiding two other bugs, one in the implementation in #411 and one I've not yet figured out.

For the unit tests

1. `test_compile_target_device_opset` was running successfully, but not actually testing the behavior because the input circuit gate was already in the basis of the target device. When I switched the input circuit to have a non-supported gate, the test would fail. This was due to the line below, which was not using the newly introduced `target_basis` property. Switching to that now makes the test pass
https://github.com/unitaryfoundation/ucc/blob/4d9a11675319aaf048b0c3137f56adbdc520e03a/ucc/compile.py#L63

2. `test_compile_target_device_coupling_map` is now failing, as the compiled circuit does not satisfy the coupling map. I've not dug in to figure out why yet.

### DIFF
--- a/ucc/tests/test_compile.py
+++ b/ucc/tests/test_compile.py
@@ -130,39 +130,42 @@ def test_custom_pass():
     post_compiler_circuit = compile(cirq_circuit, custom_passes=[HtoX()])
     assert_same_circuits(post_compiler_circuit, CirqCircuit(X(qubit)))
 
-    def test_compile_target_device_opset():
-        circuit = QiskitCircuit(3)
-        circuit.cx(0, 1)
-        circuit.cx(0, 2)
 
-        # Create a simple target that does not have direct CX between 0 and 2
-        t = Mybackend().target
-        result_circuit = compile(
-            circuit, return_format="original", target_device=t
-        )
-        # Check that the gates in the final circuit are all supported on the target device
-        assert set(op.name for op in result_circuit).issubset(
-            t.operation_names
-        )
+def test_compile_target_device_opset():
+    circuit = QiskitCircuit(3)
+    circuit.cz(0, 1)
+    circuit.cz(0, 2)
+    t = Mybackend().target
 
-    def test_compile_target_device_coupling_map():
-        circuit = QiskitCircuit(3)
-        circuit.cx(0, 1)
-        circuit.cx(0, 2)
+    # Check that the gates in the original circuit are not support by the target
+    # to ensure this isn't a trival check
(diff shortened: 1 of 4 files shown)


ucc/transpilers/ucc_defaults.py
```diff
@@ -50,8 +50,8 @@ def __init__(
                 target_device (qiskit.transpiler.Target): (Optional) The target device to compile the circuit for
         """
         self.pass_manager = PassManager()
-        self.target_device = target_device
-        self.target_gateset = target_gateset
+        self._target_device = target_device
```
jordandsullivan, line 53: Why are these now private attributes?

ucc/tests/test_compile.py
```diff
@@ -130,39 +130,42 @@ def run(self, dag):
...
-        result_circuit = compile(
-            circuit, return_format="original", target_device=t
-        )
-        # Check that the compiled circuit respects the coupling map of the target device
-        analysis_pass = CheckMap(
-            t.build_coupling_map(), property_set_field="check_map"
-        )
+    # Check that the gates in the original circuit are not support by the target
+    # to ensure this isn't a trival check
+    assert set(op.name for op in circuit).issubset(t.operation_names) is False
```
jordandsullivan, line 142: I don't think we want to _require_ that the gates in the original circuit are unsupported. I would Turn this into an if statement that only performs the check if the condition is False.

ucc/tests/test_compile.py
```diff
@@ -130,39 +130,42 @@ def run(self, dag):
...
+        circuit, return_format="original", target_device=t
+    )
+    # Check that the compiled circuit respects the coupling map of the target device
+    analysis_pass = CheckMap(
+        t.build_coupling_map(), property_set_field="check_map"
+    )
+
+    dag = circuit_to_dag(result_circuit)
+    analysis_pass.run(dag)
+    assert analysis_pass.property_set["check_map"]
```
jordandsullivan, line 168: The map passes failing I don't have an intuition about. @Misty-W any chance you happen to have an idea what may be causing these to fail?

jordandsullivan (commented):
Good catch on the indentation @bachase , couple questions and comments.

ucc/transpilers/ucc_defaults.py
bachase: Primarily was to see if any we had any others uses of `target_gateset` instead of `target_basis`. Happy to revert now, but it will be confusing to know when to use either of them.

PR #502, opened 2025-08-29
fix(aqc): compare and log fidelity using magnitude (np.abs)

ddri:
Fidelity from np.vdot is complex; comparisons to float thresholds and logs should use magnitude. This PR contains only the fidelity fix, cherry-picked from #501, without license/TODO changes.\n\nChanges:\n- approx_compile: compare np.abs(fidelity) to threshold; log magnitude.\n- MPS sequential: early-stop and logs use np.abs(fidelity).\n- qmprs compiler: logs use np.abs(fidelity).\n\nRationale:\n- Fidelity is |⟨ψ|ϕ⟩| in [0,1] and invariant to global phase. Comparing the complex value directly can misclassify high-fidelity states with phase.\n\nNo behavior changes for real-overlap cases; fixes misclassification/formatting for complex overlaps.

### DIFF
--- a/ucc/transpilers/aqc/mps_sequential.py
+++ b/ucc/transpilers/aqc/mps_sequential.py
@@ -220,18 +220,18 @@ class Sequential:
 
             fidelity = np.vdot(disentangled_mps.to_dense(), zero_state)
 
-            if fidelity >= self.max_fidelity_threshold:
+            if np.abs(fidelity) >= self.max_fidelity_threshold:
                 # If the disentangled MPS is close enough to the zero state,
                 # we can stop the disentanglement process
                 logger.info(
-                    f"Reached target fidelity {fidelity}. "
+                    f"Reached target fidelity {np.abs(fidelity):.4f}. "
                     f"{layer_index + 1} layers used."
                 )
                 break
 
         if layer_index == max_num_layers - 1:
             logger.info(
-                f"Reached fidelity {fidelity} with "
+                f"Reached fidelity {np.abs(fidelity):.4f} with "
                 f"maximum number of layers {max_num_layers}."
             )
 
@@ -308,7 +308,7 @@ class Sequential:
 
         fidelity = np.vdot(Statevector(circuit).data, statevector)
         logger.info(
-            f"Fidelity: {fidelity:.4f}, "
+            f"Fidelity: {np.abs(fidelity):.4f}, "
             f"Number of qubits: {num_qubits}, "
             f"Number of layers: {max_num_layers}, "
         )
--- a/ucc/transpilers/aqc/qmprs_compiler.py
+++ b/ucc/transpilers/aqc/qmprs_compiler.py
@@ -88,7 +88,7 @@ class QmprsCompiler:
 
         fidelity = np.vdot(circuit.get_statevector(), statevector)
         logger.info(
-            f"Fidelity: {fidelity:.4f}, "
+            f"Fidelity: {np.abs(fidelity):.4f}, "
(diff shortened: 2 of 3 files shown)


bachase:
@ACE07-Sev FYI in case you'd like to review.

ACE07-Sev:
Copy copy. I'll have a look soon.

ACE07-Sev:
Greetings there,

Hope all are well. Apologies for the delay, I got stuck with work.

@ddri Thanks for the PR. So, I am aware of the fidelity metric being a bit wonky, but the reason I kept it that way was a choice iirc. Through the past year I experimented with different stop criteria and that was what I believe gave the best result.

I usually keep a notepad for these odd implementation choices but couldn't find my notes for this at the moment. Now, I *think* we can for now merge this (it's a minimal change, I have written it down, and it shouldn't break anything, at worst it could affect fidelity).

@bachase I think that's all for me. I'll redo the experiments and see if there's a noticeable difference. My goal with this was to push explicitly towards 1+0j, but this should be fine too. I did use this for a while, but I changed it, and I recall there being a good reason for it. I will make a PR if I confirm it.

@ddri by the way, if you can show an example of your approach producing better results compared the current, that would be very helpful for my check.

ACE07-Sev:
> recovering from being sick, but feeling better every day. 
Sorry to hear that, glad you are feeling better. Hopefully we'll see you in the next UCC meeting :) .

jordandsullivan:
Hi @ACE07-Sev -- do you think you could finish up this PR in the next week or so?

ACE07-Sev:
Sure thing.

ucc/transpilers/aqc/mps_sequential.py
```diff
@@ -308,7 +308,7 @@ def __call__(
 
         fidelity = np.vdot(Statevector(circuit).data, statevector)
```
ACE07-Sev: Change this to
```py
fidelity = np.abs(
    np.vdot(Statevector(circuit).data, statevector)
) ** 2
fidelity = float(np.real(fidelity))
```

ucc/transpilers/aqc/qmprs_compiler.py
```diff
@@ -88,7 +88,7 @@ def __call__(self, statevector: NDArray[np.complex128]) -> QuantumCircuit:
 
         fidelity = np.vdot(circuit.get_statevector(), statevector)
         logger.info(
-            f"Fidelity: {fidelity:.4f}, "
+            f"Fidelity: {np.abs(fidelity):.4f}, "
```
ACE07-Sev: Undo this.

ucc/transpilers/aqc/qmprs_compiler.py
```diff
@@ -88,7 +88,7 @@ def __call__(self, statevector: NDArray[np.complex128]) -> QuantumCircuit:
 
         fidelity = np.vdot(circuit.get_statevector(), statevector)
```
ACE07-Sev: Change this to
```py
fidelity = np.abs(
    np.vdot(circuit.get_statevector(), statevector)
) ** 2
fidelity = float(np.real(fidelity))
```

ACE07-Sev:
@ddri After the amends, please benchmark with random states over different number of qubits and see if the performance is identical or not. Let me know if it consistently becomes lower after change.

@jordandsullivan I added the review notes. Because of my RAM limitation it may take a bit until I personally do the benchmarking with the new amends but should be fine.

Apologies again for the delay.

PR #504, opened 2025-08-29
fix(compile): use resolved target_basis for pre-translation basis_gates

### LINKED ISSUE #503: Pre-translation uses raw target_gateset instead of resolved target_basis
Describe the bug

The initial basis-translation step in `ucc.compile` uses `ucc_default1.target_gateset` (raw input/possibly None) instead of the resolved `ucc_default1.target_basis` (device ops → user gateset → default). This can ignore the target device’s operation set or the default gateset during the pre-translation, leading to inconsistent or suboptimal results.

```
basis_translated_circuit = qiskit_transpile(
    qiskit_circuit,
    basis_gates=ucc_default1.target_gateset,
    optimization_level=0,
    )
```
The comment says “Translate into the target device gateset first,” but this code path won’t use device ops when `target_gateset` is None. The pass manager later relies on `self.target_basis`, so this appears to be somewhat inconsistent? Is this intentional?

**To Reproduce**

1. Use a target device (e.g., `t = Mybackend().target)` and call:

```
from ucc import compile
   compiled = compile(circuit, target_device=t)
```

2. Observe the pre-translation step is configured with `basis_gates=ucc_default1.target_gateset` (None when no explicit gateset provided), instead of using `target_basis` that resolves to `t.operation_names`.
3. This can produce a basis translation not aligned with the device or default basis expected by later passes.

**Expected behaviour**

- The pre-translation should respect the resolved target basis (device operations if provided; else user-specified gateset; else default).
- Replace `basis_gates=ucc_default1.target_gateset` with `basis_gates=ucc_default1.target_basis`.

**Additional context**

- Proposed fix:
```
basis_translated_circuit = qiskit_transpile(
    qiskit_circuit,
    basis_gates=ucc_default1.target_basis,  # uses device ops → user gateset → default
    optimization_level=0,
)
```
- This aligns pre-translation with the downstream pass manager’s assumptions and the inline comment’s intent.

ddri:
Pre-translation currently uses the raw target_gateset (possibly None), which can ignore the device opset or default gates. This change switches to the resolved target_basis (device → user → default) to align with the pass manager and the intent described in comments.

Fixes #503.

### DIFF
--- a/ucc/compile.py
+++ b/ucc/compile.py
@@ -20,7 +20,7 @@ if current_major != REQUIRED_MAJOR or not (
 ):
     warnings.warn(
         f"Warning: This package is designed for Python {REQUIRED_MAJOR}.{MINOR_VERSION_MIN}-{REQUIRED_MAJOR}.{MINOR_VERSION_MAX}. "
-        f"You are using Python) {current_major}.{current_minor}."
+        f"You are using Python {current_major}.{current_minor}."
     )
 supported_circuit_formats = ConversionGraph().nodes()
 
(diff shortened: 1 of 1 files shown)


bachase:
@ddri -- thanks for this contribution. I noticed this as well while debugging another issue. Would you be open to incorporating the feedback and fixing the unit test from the code in #497 ?

PR #505, opened 2025-08-30
fix(transpile): enforce gate direction after routing; add Qiskit 2.x {cx,u} gateset test

ddri:
Adds a direction-correction pass after routing (GateDirection on Qiskit 2.x; CXDirection fallback) to address coupling-map failures with directed CX edges. Also adds a Qiskit-only test asserting that target_gateset={cx,u} is respected when return_format="qiskit".\n\nRefs: #497


ddri:
Disregard this. I should be making this PR on existing issues.

PR #517, opened 2025-09-08
updated readme with a table

Qubit1718:
Created a table for want to know more section


jordandsullivan:
Thanks for adding a fix @Qubit1718 ! It looks pretty good to me, but I wonder we should change the order or have fewer entries in the table, or otherwise make it a bit more visually appealing. Maybe adding emojis on the left?  :thinking: I'll ask @natestemen for a second opinion here.

Qubit1718:
Hi @jordandsullivan @bachase @natestemen , 

Thanks for pointing out your suggestions. I addressed them all. Please review them at your convenience.

Issue #491, opened 2025-08-20
Test updates to qiskit_ionq backend


bachase:
https://github.com/qiskit-community/qiskit-ionq/blob/backend-memory/qiskit_ionq/rewrite_rules.py has some work in progress passes meant to improve compiled performance for the IonQ native gates. That branch also includes updates for Qiskit 2.0 compatibility. This tasks involves testing those updates and their performance building on the notebook in https://github.com/unitaryfoundation/research/pull/32.

(Hat tip to @splch!)

bachase:
Trying to do this gives the error

>Providing non-standard gates (ms) through the ``basis_gates``

on https://github.com/unitaryfoundation/ucc/blob/f4eec0f127ebd0909adde3731b1ad955a9cc8a2d/ucc/compile.py#L76-L79

I think this is another case where we are mixing Qiskit idioms with UCC idioms. It looks like when there are non-standard basis gates (like those defined in the IonQ backend), we should be passing in the target device instead of just the basis gates. 

Have a quickfix for that, but digging a little further to see that this works end to end.

jordandsullivan:
I would prefer that we keep digging on this in UCC until we get to the bottom of it. I'll do my part with #529, and happy to have a pair programming session to understand the specifics of the problem.

If in the meantime we need to mark this as a known bug and use Qiskit defaults in the ion Q case, that's fine, but IMO we should make this our top priority to fix, as running well on different device types is in our core mission. 

As part of our plan to go back and deep dive into our UCC defaults, let's understand which passes specifically are failing to respect device-specific gates or coupling maps.

jordandsullivan:
I see your perspective. It's true the original set of passes were meant to work well in general, not with a specific device in mind. If you see a clearer path through switching to Qiskit defaults and doing device-specific customization and adding specific additional passes from there, I'm alright with that strategy.

bachase:
Focusing the qiskit + ionq_backend direction, for all optimization levels running on a random circuit, there are runs of `gpi` or `gpi2` single qubit gates which should be possible to merge. I'm guessing since these are custom gates, qiskit doesn't know how to combine them using its normal passes, so will look into what it would take to do so.

<img width="743" height="304" alt="Image" src="https://github.com/user-attachments/assets/471d2728-83ae-4c52-956d-0c8bfb9dd3d9" />

bachase:
[...]

Issue #490, opened 2025-08-20
Investigate using tket to optimize for IonQ devices


bachase:
To complement the findings in #461, this issue is to investigate cross-compiling via pytket with an appropriately defined IonQ backend https://docs.quantinuum.com/tket/extensions/pytket-quantinuum/index.html#default-compilation (https://github.com/CQCL/pytket-ionq is now archived). Given tket is built by Quantinuum, we want to see if it has better out-of-the box performance. This comparison would be added as another compiler in the notebook introduced in https://github.com/unitaryfoundation/research/pull/32.

bachase:
I'm marking this as nice to have, as the progress in #491 and the pending tket2 migration make this less a priority IMHO

PR #529, opened 2025-09-12
Discovered source of test failure:  `VF2PostLayout` in add_map_passes…

jordandsullivan:
… is not respecting the passed target_device layout.


jordandsullivan:
> You'd either have to manually run https://github.com/unitaryfoundation/ucc-bench/blob/main/benchmarks/layout_benchmarks.toml before and after, or we can consider enabling those in the set that runs automatically.

I'm going to run the benchmarks with and without this VF2PostLayout called and see the impact before deciding whether to remove it.

jordandsullivan:
Waiting on #491 and #490 to decide whether to proceed here.

PR #541, opened 2025-09-22
Update ucc API to accept Backends rather than Targets

jordandsullivan:
This should make things easier when we check the backend type to choose compiler passes and for passing more hardware info (e.g. specific qubit fidelities) lower in the compiler stack

### DIFF
--- a/ucc/transpilers/ucc_defaults.py
+++ b/ucc/transpilers/ucc_defaults.py
@@ -10,10 +10,10 @@ except ImportError:
     def default_num_processes():
         return CPU_COUNT
 
+
 from qiskit.providers import Backend
 from qiskit.transpiler import PassManager
 from qiskit import user_config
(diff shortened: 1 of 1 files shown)


ucc/transpilers/ucc_defaults.py
```diff
@@ -36,38 +36,48 @@ def default_num_processes():
 
 
 class UCCDefault1:
+    DEFAULT_GATESET = {"cx", "rz", "rx", "ry", "h"}
+
     def __init__(
         self,
         local_iterations: int = 1,
-        target_device: Optional[Target] = None,
+        target_backend: Optional[Target] = None,
```
bachase, line 44: Should the type annotation change to `Optional[qiskit.providers.Backend]`?

bachase (approved):
Small nit on the type annotation but otherwise 👍

